PR #1, opened 2016-10-17
typo fix in README

stevector:


greg-1-anderson:
Thank you for fixing the same bug twice -- I forgot to pull before squashing after merging through GitHub.

PR #7, opened 2016-11-02
Update to Drupal 8.2.2 and Drupal Console 1.0.0-rc7

greg-1-anderson:


greg-1-anderson:
This works, but there is a minor bug in Drupal Console 1.0.0-rc7 that emits an innocuous warning on every command. I'm going to wait for 1.0.0-rc8 before merging.

PR #11, opened 2016-11-06
Double quotes in README file are wrong type

johnennewdeeson:
Changed the type of the double quotes in the README file so that it can be copied and pasted into the terminal without errors.


greg-1-anderson:
Thank you.

PR #14, opened 2016-11-17
(WIP) Splitting out composer.pantheon.json

stevector:
This is a Work-in-Progress pull request to split out a `composer.pantheon.json` from the root `composer.json`. The files are merged with composer-merge-plugin.

Separating these files should provide these benefits:

- Make it clear to site owners which dependencies were added by Pantheon (those in `composer.pantheon.json` and which were added specific to the site. The conceptual split is similar to `settings.php` and `settings.pantheon.php`
- Make for easier updating of shared code. This pattern should also be beneficial to organization using Pantheon custom upstreams which use git merging to update shared code. For instance, a university/agency with many similar sites could add a `composer.acme.json` in the root directory or even in a profile.
- The `extras` key can also be merged which allows important config (that's unlikely to vary per site) like `installer-paths` and `drupal-scaffold` to be shared.

I wish the `scripts` key could be merged since we're starting to make heavy use of it. I've commented on an issue in composer-merge-plugin with an idea for how to merge the `scripts` key: https://github.com/wikimedia/composer-merge-plugin/issues/68

Some open questions:

- Is this change worth merging without shared `scripts` being shared? The merging of `require` and `require-dev` is nice but doesn't do anything that couldn't be done with `composer require pantheon-systems/some-hypothetical-shared-project`.
- Is the increased complexity of multiple composer files worth the clarity of separating shared config from site specific config? I think it is, especially when composer-merge-plugin is already used in Drupal core for the same concept with `/core/composer.json`.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -13,6 +13,13 @@ dependencies:
   cache_directories:
     - ~/.composer/cache
   pre:
+    # Tests on this pull request are currently failing. I suspect that is because
+    # with a composer lock file, all the dependencies can be downloaded
+    # immediately. That means drupal core can be downloaded before the merge
+    # plugin has processed the shared settings that say the installer path
+    # for core is web/core
+    # THIS CHANGE SHOULD NOT BE MERGED. IT IS FOR DEBUGGING THE PULL REQUEST.
+    - rm composer.lock
     # Set the PHP timezone so that Behat script does not fail.
     # Using > instead of >> will overwrite the file and disable xdebug.
     # xdebug makes composer slower.
--- a/composer.json
+++ b/composer.json
@@ -10,27 +10,7 @@
     }
   ],
   "require": {
-    "composer/installers": "^1.0.20",
-    "drupal-composer/drupal-scaffold": "^2.0.1",
-    "cweagans/composer-patches": "~1.0",
-
-    "drupal/core": "~8",
-
-    "drupal/console": "^1.0.0-rc8",
-    "drush/drush": "~8"
-  },
-  "require-dev": {
-    "mikey179/vfsstream": "~1.2",
-    "behat/behat": "3.*",
-    "behat/mink": "~1.7",
-    "behat/mink-extension": "*",
-    "behat/mink-goutte-driver": "~1.2",
-    "jcalderonzumba/gastonjs": "~1.0.2",
-    "jcalderonzumba/mink-phantomjs-driver": "~0.3.1",
-    "drupal/drupal-extension": "*",
-    "drush-ops/behat-drush-endpoint": "*",
-    "phpunit/phpunit": "~4.8",
-    "symfony/css-selector": "~2.8"
+    "wikimedia/composer-merge-plugin": "~1.3"
   },
   "conflict": {
       "drupal/drupal": "*"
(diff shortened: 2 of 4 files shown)


stevector:
I'm looking into the test fails.

stevector:
The only commit in this PR that passed tests was the one that had all of the `extras` key in `composer.json` instead of `composer.pantheon.json`.

There are a different patterns of fail:

https://circleci.com/gh/pantheon-systems/example-drops-8-composer/60
When `extras` is in `composer.pantheon.json` and a `composer.lock` file is present, it seems to me that Drupal core gets downloaded and placed in `/core` instead of `/web/core`. Presumably because the download has happened before `extras` were merged? I'm not certain of that. Even in build 60 here, the composer-merge-plugin was downloaded before core.

https://circleci.com/gh/pantheon-systems/example-drops-8-composer/62
In this build I tried deleting the `composer.lock` file before running `composer install`. Doing so allowed the composer installation to complete. But composer `composer drupal-scaffold` then failed because it depends on `DrupalComposer\DrupalScaffold\Handler::getDrupalCoreVersion()` reading the version of Drupal Core from `composer.json`. It seems composer-merge-plugin is not processed here. I suspect the change I've suggested in composer-merge-plugin itself might address this problem by processing the merge in far more (all?) cases: https://github.com/wikimedia/composer-merge-plugin/issues/68

https://circleci.com/gh/pantheon-systems/example-drops-8-composer/63
https://circleci.com/gh/pantheon-systems/example-drops-8-composer/64
Moving Drupal Core back to `composer.json` brought back the same fail that was seen in Circle Build 60.

@greg-1-anderson, this problems could be fixable with much more testing/tweaking. Do you think this change is worthwhile if the only elements split into `composer.pantheon.json` are `require` and 'require-dev'?

greg-1-anderson:
Splitting out `require` into a separate composer.json file is a very valuable addition, IMO.

Regarding the behavior of `extras`, try including the composer-merge-plugin before composer-installers.  I bet that composer-installers reads and caches extras as soon as it is included.

greg-1-anderson:
Hm, still not relocating things correctly.  😞

stevector:
Yeah, flipping them did not make a difference. @greg-1-anderson what's your level of familiarity with the mechanics of Composer plugins? Do you think what I suggested over in https://github.com/wikimedia/composer-merge-plugin/issues/68 would help with `extras`?

greg-1-anderson:
Possibly. Might take stepping through the code (or at least careful examination) to really know what's going on vis-a-vis the composer-merge-plugins `extra` handling and the composer installers plugin.

greg-1-anderson:
I moved the `repositories`, `autoload` and 'conflict` directives over successfully.  This leaves us with just `minimum-stability`/`prefer-stable`, `scripts` and `extras` among the things that we'd like to move that don't work.

It's a little surprising that `minimum-stability` does not work, as this directive has been moved to `core/composer.json` in Drupal 8, and presumably works there.

It is perhaps a little odd to have the autoload directive used by the scripts to be in the other composer.json, but I put it there to be forward-looking to a time when we might be able to move the scripts over too. It's fine by me to move this back for consistency, if you prefer.

I suspect that it might be low-effort to put a PR in wikimedia/composer-merge-plugin to support relocation of the script element. I think this would be a win for us. I may look into trying what you suggested if I have time; I'd also encourage you to continue in this direction if you feel so inclined. The other idea I had about scripts is that we could move the Pantheon composer.json file to `pantheon/composer.json`, and then define a script called `pantheon` that just did a `cd pantheon && composer`. Didn't try it, but this should allow you to run `composer pantheon my-pantheon-script` to run a script in the `pantheon/composer.json` file. That wouldn't be horrible, but better to fix `script` relocation, I think.

I am less concerned with the remaining `extras` section.  I think this is not likely to change much, and I suspect it might be a larger effort to get it working correctly, so I am content to leave it. It would be nice to be able to move it if it did happen to get fixed, though.

I'd be fine with squash-merging this as-is, if you'd like; or, we could keep it open until the `scripts` issue is resolved.

PR #16, opened 2016-12-06
Merge multidev env to dev after feature branch merged to master

greg-1-anderson:
Add a merge-pantheon-multidev script to merge the tested multidev into the dev environment after any branch is merged to 'master' (after the master branch tests pass). For testing, though, we will merge all test environments back to dev.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -33,3 +33,10 @@ test:
   override:
     - terminus composer drupal-unit-tests
     - cd tests && ./scripts/run-behat
+
+deployment:
+  build-assets:
+    branch: master
+    commands:
+      - tests/scripts/merge-pantheon-multidev
+      - terminus drush updatedb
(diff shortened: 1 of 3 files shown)


stevector:
@greg-1-anderson, is the idea here that when merging to master, a mutlidev already exists where the some have been run? If so, the merge could be as simple as `terminus site merge-to-dev`, right?

greg-1-anderson:
stevector I don't think that would always work due to potential merge conflicts in the `vendor` directory.

greg-1-anderson:
Merging to test deployment via this technique.

Need to be a little creative to test `terminus site merge-to-dev` by making two PRs off of the same commit that make changes that do not conflict with each other, but cause conflicts in `vendor`. Or read the Terminus source code and see if it does anything special, or just runs `git merge`. The later (with all of the protection of `git`) is probably what's wanted most of the time.

greg-1-anderson:
I don't think that `terminus site merge-to-dev` will work (`terminus site merge-to-dev` just ends up running `git merge`), but perhaps `git merge -X theirs` would be more efficient (and less "magical") than the current implementation of replacing one tree with another.

PR #18, opened 2016-12-07
Correct updatedb calls in circle.yml.

greg-1-anderson:
updatedb call missing after pushing to CI environment, and targeted wrong multidev after merging to dev environment.


greg-1-anderson:
Merged manually.

PR #19, opened 2016-12-07
Improve style of 'visit site' link posted by CI bot.

greg-1-anderson:


derimagia:
I mention in another post that I'm posting a lot of this to gitlab so I went through a lot of the code for it. Minor but since this pull request already exists, is there a reason why the comment is exported?

greg-1-anderson:
Do you mean why `export comment=` instead of just `comment=`?  No reason, it was just there like that. Don't know why I didn't bother to take it out. Might as well do that.

greg-1-anderson:
... and now the tests fail until I have a tag for 8.2.4 in drops-8.  Or pin to 8.2.3 in composer.json. Sigh.

derimagia:
Yeah... I got the same thing on my work I'm doing.

greg-1-anderson:
Limitation of the way I'm using drupal-scaffold.  We take the best tag from drupal/core, and expect that drops-8 will have the same tag.  I'll do the update of drops-8 shortly.

greg-1-anderson:
@derimagia Where are you doing your GitLab work?  If it's not too different, it might make sense to integrate.  Maybe after switching the scripts from Bash to Robo, so that the repository-specific parts can be factored out.

derimagia:
Right now at gitlab private repo, but I was planning to move it over to a pull request on github

Robo sounds like a great idea, once I test what I have further I'll take a look at it

PR #24, opened 2016-12-09
Gitlab CI Implementation

derimagia:
Some comments about this on #19. Want some feedback on it.

1. Added gitlab-ci implementation /  .gitlab-ci.yml
2. Uses Environments on Circle CI
3. Added Gitlab Comments similar to Github Comments, but with Environments I'm second guessing if this was needed.
4. Changed run-behat to not need to be in a specific directory
5. Added TERMINUS_LABEL variable for commit messages
6. Defaulted variables in scripts to reasonable defaults, checking both variables for CircleCI and the "Generic" Version of the variables for Gitlab CI

This does not have work with Robo in it 

One decision I probably will make for this is to not have commit specific environments but change it to branch-specific environments. The script that exists doesn't delete old commit-branches, I noticed. It was only deleting the multi-dev so there were a lot of branches left on Pantheon.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,74 @@
+image: composer
+
+# Default Variables
+variables:
+  TERMINUS_SITE: mg-ci-example-d8
+  TERMINUS_ENV: ci-$CI_PIPELINE_ID
+  TERMINUS_ENV_LABEL: CI-$CI_PIPELINE_ID
+  MULTIDEV_DELETE_PATTERN: ^ci-
+  CGR_BASE_DIR: /composer/.global
+  CGR_BIN_DIR: /composer/vendor/bin
+
+# Default Cache Directories
+cache:
+  paths:
+    - /composer/cache
+
+stages:
+- build
+- test
+- deploy
+
+before_script:
+  # See https://docs.gitlab.com/ee/ci/ssh_keys/README.html
+  - eval $(ssh-agent -s) && ssh-add <(echo "$SSH_PRIVATE_KEY")
+  # Avoid ssh prompting when connecting to new ssh hosts
+  - mkdir -p $HOME/.ssh && echo "StrictHostKeyChecking no" >> "$HOME/.ssh/config"
+  # Git Config
+  - git config --global user.email "$GITLAB_USER_EMAIL"
+  - git config --global user.name "Gitlab CI"
+  # Composer and Terminus Setup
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
Thanks for the improvements!

Since the scripts themselves don't really have much Gitlab-specific branching (just comment posting), converting to Robo or some other higher-level scripting language is not necessary for this PR (but still a nice to have, as a separate independent effort).

Regarding feature branches vs build-number branches, the problem with the former is that your tests will conflict and fail if multiple commits are made to a branch before the running tests can complete. Unless you add some sort of build semaphore on CircleCI, the later build will ruin the earlier build. Perhaps this is a limitation that folks would be willing to live with. Using a single feature branch also has the advantage of having a more stable test environment, e.g. if using the test environment to showcase a feature to stakeholders through multiple revisions.

One thought I had in this area was to vary the multidev environment name depending on the naming conventions of the PR branch.  If the branch starts with 'feature-', then the remaining portion of the branch name would be used in the environment name.  e.g. the `feature-foo` branch would create a multidev environment `ci-foo` (or something else not eligible for deletion by delete-old-multidevs -- feature branches would clean up on merge). If the branch did not start with `feature`, then the build number would be used as usual. This would put the behavior of the environment creation in the hands of the user.

Regarding branch cleanup, yes, the script should definitely delete the branch when cleaning up old multidevs.  I just need to add `--remove-branch` to the Terminus command when deleting the multidev. I'll do that momentarily.

derimagia:
Another option would be to have 1 multi-dev per feature branch and then every commit in the branch would get it's own environment with the test running on that one. The branch multi-dev can be updated with code like usual, but that was you have both.

To deal with a semaphore could also use a file lock could be used in the files directory of the site. I do like the idea of options, though, even if it was just a variable

derimagia:
How did you want to do tests for the Gitlab section? Have any thoughts/interests? Maybe just sync the repo to a gitlab repo with a different terminus_env?

greg-1-anderson:
If you want to maintain the gitlab repo, that would be appreciated. Otherwise, I am satisfied to leave the gitlab file as an untested example.

greg-1-anderson:
Did some refactoring of the circle.yml file. Earlier I made some passing effort at also  updating the gitlab CI, but it is probably not working now.

derimagia:
Just as an update - still plan on doing this. @rvtraveller was taking a look and I believe there's just a few tweaks to do

PR #30, opened 2016-12-28
Relocate private files outside of the document root.

greg-1-anderson:
Included .htaccess files for nominal protection should the site be moved off of Pantheon. This ostensibly isn't necessary for directories outside of the document root, but the `config` folder already has one. :man_shrugging:

### DIFF
--- a/private/.htaccess
+++ b/private/.htaccess
@@ -1,0 +1,23 @@
+# Deny all requests from Apache 2.4+.
+<IfModule mod_authz_core.c>
+  Require all denied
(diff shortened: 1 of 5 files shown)


greg-1-anderson:
I made this to experiment with archive-dump files that contain private files outside of the document root, but Drush does not support those.

Neither can this PR be merged, as Pantheon only supports file sharing between multiple app servers through the git repository and the files directory. In this PR, the `private` files are in neither location, so this will not work.

PR #15, opened 2016-11-27
Add Lightning and the Acquia Lightning ScriptHandler.

greg-1-anderson:
This PR is not intended to be merged.  This is an experiment to combine example-drops-8-composer with Acquia Lightning.

At the moment this is not working; packagist.drupal.org seems to place the entire distribution at profiles/lightning.  See https://github.com/acquia/lightning-project/issues/22

PR #33, opened 2017-01-01
Lightning fist 2

greg-1-anderson:
Continuation of #15


greg-1-anderson:
This PR uses Lightning v2, which is based on packagist.drupal.org, just as drupal-composer/drupal-project is. Includes a couple of simple passing tests, including an installer test.

PR #35, opened 2017-01-04
Use composer create-project in readme

stevector:
@greg-1-anderson, since your forthcoming blog post uses `composer create-project`, I think the readme here should too.


greg-1-anderson:
Good idea; thanks. I do plan on updating the README to include the PR workflow setup instructions as well. The blog is now being converted into a guide in Pantheon documentation. Should be up soon.

PR #48, opened 2017-01-17
Fix manual install

greg-1-anderson:


greg-1-anderson:
Merged via commandline

PR #50, opened 2017-01-23
Moving terminus vars out of circle.yml

stevector:
@greg-1-anderson, I think these vars should be set through the CircleCI UI https://circleci.com/gh/pantheon-systems/example-drops-8-composer/edit#env-vars but I have not done so yet (and that will probably cause a test fail here)

I want to minimize the number of file changes that have to be done when making a new copy.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -6,6 +6,29 @@ machine:
     # https://circleci.com/docs/build-image-trusty/#php
     version: 7.0.11
   environment:
+    # This repository requires environmental variables that are likely to change
+    # as it is copied for different websites. Those variables can be defined
+    # in the CircleCI UI which can keep the values of variables hidden.
+    # That functionality is important for variables like these:
+    ### The password to be used for Drupal's user 1 account.
+    # ADMIN_PASSWORD: xxxxxxxxxxxxxxx
+    ### Generate from https://dashboard.pantheon.io/users/#account/tokens/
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
It is not necessary to change the default values in circle.yml -- if you add the same environment variable in the Circle CI Web UI, it will override the default.

However, I agree that might be confusing, so this seems like a good idea. However, I think this PR should also include a comment listing all of the variables that must be defined in the Circle CI UI here, so they are all in one place. Or maybe that one place should only be in the README.

Compare with the documentation in the [Terminus Build Tools Plugin](https://github.com/pantheon-systems/terminus-build-tools-plugin)

greg-1-anderson:
I added the environment variables removed in this PR to the Circle CI project environment variables through the web ui. I'll re-run the tests.

greg-1-anderson:
+1

Feel free to merge this (with or without docs updates) when you are ready.

stevector:
@greg-1-anderson, what do you think of these explanatory comments?

stevector:
Should I move ADMIN_EMAIL to the sensitive list with ADMIN_PASSWORD?

greg-1-anderson:
+1 LGTM. I think the reader can decide if the various email addresses are "sensitive" or not. I'm indifferent to whether these are moved, or left where they are. They're fine as-is.

PR #51, opened 2017-01-24
Fix drush Behat Step

stevector:
@greg-1-anderson, with single quotes, the `$site` and `$env` variables are not processed correctly.


greg-1-anderson:
You're right. Probably broke during the Terminus 1.x upgrade; I guess we're not running Drush commands from Behat right now.

PR #53, opened 2017-01-27
Use rvtraveller/qs-composer-installer 

greg-1-anderson:
(my fork) instead of my fork of composer/installers.


greg-1-anderson:
Update to use rvtraveller/qs-composer-installer directly now that my PR has been merged.

PR #56, opened 2017-02-04
Add pantheon-systems/terminus as a dev dependency

arshad:
.+.
        .+?:
         .+??.
           ??? .
           +???.
      +?????????=.
      .???????????.
      .????????????.

     ########### ########
     ############.#######.
     ####### ####  .......
     ######## #### #######                50 41 4E 54 48 45 4F 4E
     #########.####.######        _____________  __  ________  ____  ______
     ######  ...                 /_  __/ __/ _ \/  |/  /  _/ |/ / / / / __/
     #######.??.##########        / / / _// , _/ /|_/ // //    / /_/ /\ \
     #######~+??.#########       /_/ /___/_/|_/_/  /_/___/_/|_/\____/___/
     ########.??..
     #########.??.#######.
     #########.+?? ######.
               .+?.
         .????????????.
           +??????????,
            .????++++++.
              ????.
              .???,
               .~??.
                 .??
                  .?,.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,8 @@
     "drush-ops/behat-drush-endpoint": "^0.0.4",
     "pantheon-systems/quicksilver-pushback": "~1",
     "phpunit/phpunit": "^4.8",
-    "symfony/css-selector": "^2.8"
+    "symfony/css-selector": "^2.8",
+    "pantheon-systems/terminus": "^1.0"
   },
   "conflict": {
       "drupal/drupal": "*"
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#composer-lock-the-lock-file",
         "This file is @generated automatically"
     ],
-    "content-hash": "45d1a1ee68f6af758ac8c5d4546c9f1c",
+    "content-hash": "4ac98073cdfacd9524142b96343f82ac",
     "packages": [
         {
             "name": "alchemy/zippy",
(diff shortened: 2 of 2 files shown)


stevector:
@arshad, I have avoided adding Terminus as a dev dependency in our repos because there is a possibility of versioning collisions.

It also shifts the mental model of the relationship between Terminus/Pantheon and the given site. By adding Terminus to the the composer.json of a site, Terminus is literally inside of the site.

We've gotten a few questions even about running Terminus from inside a site on Pantheon. So far, we've discouraged that (though it is technically possible to my understanding) because Terminus relies on an auth token for a user. Terminus is meant to allow people (or scripts acting on their behalf) to control their accounts including their sites. Putting Terminus inside a site could let that site control other sites or the account of the user who created the site.

This question is somewhat academic since all we may be talking about here is whether Terminus installs on CircleCI through `composer.json` or `cgr`. Is there a specific benefit you are looking for?

@greg-1-anderson, what are your thoughts?

greg-1-anderson:
It is inappropriate to install Terminus as a dependency of this project. Terminus and Drupal have different requirements for their Symfony dependencies. The Terminus composer.json claims to support Symfony versions that are compatible with Drupal, and yet, if you use those, then the Terminus unit tests fail.

Terminus should continue to be installed from the Circle scripts, not as a dependency of this project.

greg-1-anderson:
Ideally, Terminus would update its composer.json so that failing builds fail to install. I started to work on this at https://github.com/pantheon-systems/terminus/pull/1593, but gave up on it. Maybe it's still worth doing.

PR #62, opened 2017-03-01
Removing unneeded sshconfig

stevector:
We have other repos that don't need this. @greg-1-anderson have you seen errors when this config isn't present?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -41,8 +41,6 @@ dependencies:
     - ~/.composer/cache
   pre:
     - echo "Begin build for $CI_LABEL${PR_LABEL:+ for }$PR_LABEL. Pantheon test environment is $TERMINUS_SITE.$TERMINUS_ENV"
-    # Avoid ssh prompting when connecting to new ssh hosts
-    - echo "StrictHostKeyChecking no" >> "$HOME/.ssh/config"
     - |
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
@stevector Without that configuration, you will sometimes (often) get stuck at a prompt to accept the credentials for the remote server. This only happens when you are running remote Drush or WP-CLI commands. I think this configuration needs to stay.

I'm going to try to re-run the test without the cache, and see if it hangs up. I suspect it only worked this time because Circle ran a previous test with the no host key checking, and cached the credentials.

greg-1-anderson:
Hm, it said `Warning: Permanently added '[appserver.pr-sshconfi.c401fd14-f745-4e51-9af2-f30b45146a0c.drush.in]:2222,[104.130.220.74]:2222' (RSA) to the list of known hosts.`, which implies that the no host ssh key configuration is active.

stevector:
@greg-1-anderson, have you ever seen a CircleCI build get hung on a prompt to accept the remote server? Our WordPress plugin and Drupal module repos call Drush and wp-cli commands from CircleCI on Pantheon. But they don't explicitly set `StrictHostKeyChecking no`

greg-1-anderson:
Maybe this config line is needed in Travis; CircleCI sets this option in `/etc/ssh/ssh_config`.

PR #65, opened 2017-03-12
Use config_installer.

greg-1-anderson:
Once configuration has been exported, use the config_installer for every subsequent install.

The previous technique (`drush site-install --config-dir` option) only works well for the `minimal` profile.


greg-1-anderson:
Merged 236bf9e

PR #66, opened 2017-03-13
Make the first build always happen on the dev environment.

greg-1-anderson:


greg-1-anderson:
Moving away from this idea.

PR #70, opened 2017-03-30
Fix wording in README.

greg-1-anderson:


stevector:
I think this PR is deprecated by https://github.com/pantheon-systems/example-drops-8-composer/pull/101

PR #91, opened 2017-05-30
Fixing typo

sukottokun:


greg-1-anderson:
This isn't a typo. `build-env:merge` is the old name for the command. `build:env:merge` is the new name for the command.  The latest version of the Terminus build tools on the 1.x branch still uses the old name as the @command name, and has an alias for the new command name. On the 2.x branch, the new command name is used for the @command annotation, and the old name is available as an alias.

PR #92, opened 2017-06-22
Update README.md with admin user info

leorawe:
Either add a line to set admin username/password at drupa install or terminus drush sitename.env -- uli to get a reset link


greg-1-anderson:
This is a helpful addition to the README. Thanks.

PR #94, opened 2017-07-11
Remove .git directories

greg-1-anderson:
Prevent git from interpreting these folders as submodules, which prevents their contents from being added to the Pantheon deployment repository.

See: https://www.drupal.org/node/2886531, item 2.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
       "url": "https://packages.drupal.org/8"
     }
   ],
-    "require": {
+  "require": {
     "composer/installers": "^1.0.20",
     "cweagans/composer-patches": "^1.0",
     "drupal-composer/drupal-scaffold": "^2.0.1",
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
As discussed with @stevector @ataylorme @populist, this will need to happen in other repositories, and ideally would be provided as a composer installer that could be `reqiure`d as needed.

drugan:
@greg-1-anderson 

Did you try to rename .git folders to something else (ex.- hidden.git) and then to exclude those in .gitignore.

I use the same workflow like yours one: pull drupalcommerce into my local repository and then pushing it with all dependencies to a server.

See more: https://www.drupal.org/node/2847898#comment-11898879

greg-1-anderson:
This is presently handled in the prepareForPantheon script in the scripthandler.

PR #95, opened 2017-07-14
Git should global ignore file permissions

stevector:
This change got https://github.com/pantheon-systems/terminus-build-tools-plugin/pull/60 back to passing @greg-1-anderson. I'm still not certain why we are getting different file permissions between the master branch generated in the Build Plugin's Circle 2.0 and the `ci-1` branch generated in the Circle 1.0 container from derivative repo. But I think it is safe, even preferable, to global ignore file permissions in git.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -40,6 +40,7 @@ dependencies:
       fi
     - git config --global user.email "$GIT_EMAIL"
     - git config --global user.name "Circle CI"
+    - git config --global core.fileMode false
   override:
     - composer global require -n "hirak/prestissimo:^0.3"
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
Is this only an issue with Circle 2.0? If so, why change the circle.yml file?

Git only records the `x` bit in the repository. We are not using that here, so it is fine to make this change. I think we should only do it where necessary, though.

greg-1-anderson:
Ah, I see. The repository is created by `build:project:create` under Circle 2.0, and then the tests run under Circle 1.0 here.

Merging.

PR #101, opened 2017-08-14
Updating readme to be more like example-wordpress-composer

stevector:
@greg-1-anderson, @ataylorme, here is a readme update we discussed a few weeks ago.

Most critically this update removes references to manual setup as well as using the repo as an upstream.


greg-1-anderson:
Yep, this is good. Thanks.

PR #102, opened 2017-08-25
Add ratelimit check after all steps of test.

greg-1-anderson:


greg-1-anderson:
Not needed any longer

PR #106, opened 2017-09-07
Require PHP 7.0 during dependency resolution, even if running 7.1+

greg-1-anderson:
Related: https://github.com/pantheon-systems/example-wordpress-composer/pull/34

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -103,6 +103,9 @@
   "config": {
     "optimize-autoloader": true,
     "preferred-install": "dist",
-    "sort-packages": true
+    "sort-packages": true,
+    "platform": {
+      "php": "7.0"
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -101,6 +101,9 @@
   "config": {
     "optimize-autoloader": true,
     "preferred-install": "dist",
-    "sort-packages": true
+    "sort-packages": true,
+    "platform": {
+      "php": "7.0"
```
stevector: I did not know `platform` was an option! @greg-1-anderson shall we merge this?

greg-1-anderson: One change of behavior here is that not only does this change allow correct operation of `composer update` when php 7.1 is installed, it also allows `composer update` to work correctly when you are running php 5.6. Adding a php constraint in the `require` section does not help with this.

I guess that could be considered an advantage. Even if we think it is a disadvantage, I think that the benefits outweigh the disadvantage, and we should merge these.

PR #120, opened 2017-11-15
Don't recommend Terminus to run composer update

dustinleblanc:
As most users tend to run out of memory when running `terminus composer site.env -- update`, we should just stop telling them to do it. This is almost guaranteed to run out of memory. We should choose a single supported way to do this as adoption is growing and it is creating confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,19 +67,7 @@ So that CircleCI will have some test to run, this repository includes a configur
 
 ## Updating your site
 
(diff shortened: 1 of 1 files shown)


greg-1-anderson:
:+1:

stevector:
I like the substance of this change. But in seeing this PR I realize that as written currently (and in the PR) we are suggesting interacting directly with the Pantheon repo. Shouldn't we be steering people toward their GitHub repo?

greg-1-anderson:
You are right -- I just read the text and thought this was in pantheon-systems/docs somewhere. For this repository, we shouldn't talk about git mode et. al.

PR #121, opened 2017-11-15
Use Terminus Build Tools plugin 2.0.0-alpha4 and later.

greg-1-anderson:
Just to see if it works. Probably should just remove and let the version in the docker container be used.


greg-1-anderson:
Back to 1.x.

PR #124, opened 2017-12-08
added drush cr after DB is updated, 

nimzie:
to be sure all paths internally are properly reconciled where files may have been moved to other locations and fixing the "The website encountered an unexpected error. Please try again later." error which was thrown. I'd moved some TPL files around and since I'm cloning from dev in my build, it fell flat.


nimzie:
@greg.1.anderson, this PR does a drush cache-rebuild after db is imported in case owner may have moved themes or modules around after initial commit.
CI is failing, let me know if you have any changes to suggest. Thank you. (first official PR on GITHUB)

PR #126, opened 2017-12-15
Two Small Documentation Tweaks

jproctor:
As I was setting up a new project and generally learning my way around, I noticed two small issues with the documentation:

- .circleci/config.yml had a bad URL for environment variables documentation

- Where web/sites/default/settings.php sets the install profile, the comment referred to tests/installer-features/installer.feature, which was removed by PR #58


greg-1-anderson:
Thanks.

PR #128, opened 2018-01-10
Contenta (Take2)

stevector:
@greg-1-anderson here's a PR on a PR. This is an update to the Contenta branch.

I've gotten it to install successfully via Build Tools with

```
terminus build:project:create pantheon-systems/example-drops-8-composer:dev-contenta--take2  contenta-take2-e  --team="ACEs"
```

Let merge this and then pick up the conversation back in #93


stevector:
@greg-1-anderson, here is the part you might be most interested in that allows for `drush site-install` to complete: https://github.com/pantheon-systems/example-drops-8-composer/blob/99c0db34292bca742b363c0a4f3f212669cac89f/web/sites/default/settings.php#L31

```
if (function_exists('drush_get_arguments')) {
  $drush_args = drush_get_arguments();
  if (!empty($drush_args[0])  && 'site-install' === $drush_args[0]) {
    $config_directories['sync'] = 'profiles/contrib/contenta_jsonapi/config/sync';
  }
}
```

greg-1-anderson:
Just two questions:

1. Why is web/libraries/dropzone checked in to the repository?
2.Why did you switch from my install hack to yours (above)? I guess mine wasn't working any longer. The way you've done it is Drush-8-specific, so installs from Drush 9 or Drupal Console would fail.

stevector:
> Why is web/libraries/dropzone checked in to the repository?

content_jsonapi_project has it fully committed as well. 
https://github.com/contentacms/contenta_jsonapi_project/tree/8.x-1.x/web/libraries/dropzone

I should have said directly that I'm treating this branch as a conglomeration of `example-drops-8-composer` and `contenta_jsonapi_project`.

That being said, I don't like the fact that dropzone is fully committed. My understanding is that it is fully committed because there is not a great way to include via composer alone. But I'd like to revisit this question with the Contenta team.

> Why did you switch from my install hack to yours (above)?

So that `terminus build:project:create pantheon-systems/example-drops-8-composer:dev-contenta--take2` could complete without error. I think the real fix will need to happen in Contenta as you suggested here https://github.com/pantheon-systems/example-drops-8-composer/pull/93#issuecomment-313920864

greg-1-anderson:
> content_jsonapi_project has it fully committed as well.

Okay. Maybe the Contenta team could consider downloading it in a post-update task.

> So that [the build] could complete without error.

Yes, that was the purpose of my hack as well. I was just wondering if there was some way to detect that installation was happening, as my old hack did, so that we don't fail in the Drush 9 / Drupal Console case. Do you have any insight into why the old hack does not work any more?

stevector:
> Do you have any insight into why the old hack does not work any more?

I thought I had tried it unsuccessfully. But I was wrong. I just tested it again and it worked so I have brought your original settings.php hack back to this branch.

broeker:
@stevector thanks for chiming in on the contenta issue I posted -- I updated my terminus plugins to latest versions, and then ran this command:

`terminus build:project:create --org="electriccitizen" --team="Electric Citizen" pantheon-systems/example-drops-8-composer:dev-contenta--take2 lawhelp`

Everything seemed to spin up correctly through the "Successfully applied `pantheon.yml` to the 'dev' environment" message but ended up with this error right after the "Install site on lawhelp.dev" step:

`Checking out files: 100% (23413/23413), done.
HEAD is now at 55ad450 Initial commit
 [notice] Install the site on the dev environment
 [notice] Install site on lawhelp.dev
 **[error]  Successfully queued converge_appserver_binding**` 

I'll give it another but let me know if you have an idea...

broeker:
@stevector I guess I should have tried twice -- looks like the error was just a random network/app problem because on attempt number the install went flawlessly and we now have a github+circle+pantheon contenta site. Next beer is on me :)

PR #132, opened 2018-02-09
Use https for behat tests

deviantintegral:
I can't think of any reason to point to http, given that even free accounts have https support.

### DIFF
--- a/tests/scripts/run-behat
+++ b/tests/scripts/run-behat
@@ -25,5 +25,5 @@ terminus aliases
 # Drush Behat driver fails without this option.
 echo "\$options['strict'] = 0;" >> ~/.drush/pantheon.aliases.drushrc.php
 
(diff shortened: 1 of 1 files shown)


deviantintegral:
> Begin build for pr-132. Pantheon test environment is .pr-132
 [notice] Logging in via machine token.
 [error]  Server error: `POST https://terminus.pantheon.io/api/authorize/machine-token` resulted in a `500 Internal Server Error` response:
"Authorization failed. Please check that your machine token is valid."

Interesting, this is from the very first step. I would have guessed that machine tokens would be in Circle environment variables, but perhaps there's a setting to omit them from builds from forks?

greg-1-anderson:
@eviantintegral The Circle environment variables contain secrets, and are therefore not applied to when testing external PRs. It was my thought to try doing a limited Pantheon-free test for external PRs, but I haven't gotten to that yet.

The change here looks safe, as the Pantheon DNS entries always have https configured.

Issue #141, opened 2018-03-20
composer.json indent size


alisonjo2786:
(hoping it's not a dupe issue; I promise I searched first!)

Y'all's composer.json file is indented with 2-space indents, but indent size for composer.json/lock files is supposed to be 4.

And yet, y'all's composer.LOCK file is 4-space-indented.

I only noticed b/c I was checking/making sure of some stuff between my composer.json in a project that was originally based on example-drops-8-composer, and before I knew it, my indents were different, and then I couldn't figure out why....... and here I am.

My initial guess is that it's b/c y'all don't have the .editorconfig that comes with Drupal core, or the one in drupal-scaffold, in this repo.

Anyway, here's my source material:
https://github.com/drupal-composer/drupal-scaffold/blob/master/.editorconfig
_(and like I said, it's in the one that lives in web/ after you install Drupal)_

alisonjo2786:
Er, wow, GitHub is fancier than I realized... didn't mean to "reference this issue" from a commit to my own repo.......

greg-1-anderson:
composer.lock is a generated file, so it's hard to get that one wrong. 😄 

I agree that we should fix the indentation of composer.json

alisonjo2786:
Oh right, good point, I think I thought too hard about that one :grin:

alisonjo2786:
Would you want to do it as a one-time fix, or with an .editorconfig file in the repo?  (I was going to offer a PR, if it'd be helpful.)

greg-1-anderson:
Both changes together would be welcome.

greg-1-anderson:
Fixed. Thanks.

PR #142, opened 2018-03-21
.editorconfig + composer.json indentation

alisonjo2786:
#141 

.editorconfig copied from https://github.com/drupal-composer/drupal-scaffold/blob/master/.editorconfig

(this is pretty basic, idk if there's a more dynamic way to include the .editorconfig file, but here 'tis!)


greg-1-anderson:
Thanks!

PR #143, opened 2018-03-24
added run step to fix permissions on scripts.

jraviotta:
When executing terminus project:build:create in a local container, scripts loose the +x bit on clone. This edit adds a run step to circleci config.yml to chmod 755 all files in /root/example_drops_8_composer/.circleci/scripts/pantheon/ & /root/example_drops_8_composer/tests/scripts/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,12 +98,6 @@ jobs:
                 command: /build-tools-ci/scripts/set-environment
 
             - run:
-                # Set permissions for scripts that get changed during clone
-                name: change script permissions
-                command: find /root/example_drops_8_composer/.circleci/scripts/pantheon/ -type f | xargs chmod 755 && \
-                         find /root/example_drops_8_composer/tests/scripts/ -type f | xargs chmod 755
-
-            - run:
                 name: install dev dependencies, build assets, etc.
                 command: ./.circleci/scripts/pantheon/01-prepare
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -98,6 +98,12 @@ jobs:
                 command: /build-tools-ci/scripts/set-environment
 
             - run:
+                # Set permissions for scripts that get changed during clone
+                name: change script permissions
+                command: cd /root/example_drops_8_composer/.circleci/scripts/pantheon/ && find . -type f | xargs chmod 755 && \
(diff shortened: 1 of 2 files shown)


greg-1-anderson:
This is a good point, although I do wonder why the tests are passing at all without +x on the scripts.

It would be better to do this at build time, though, before committing the scripts to the new git repository.

jraviotta:
@greg-1-anderson when building in a local docker container, my build_deploy_and_test tests do not pass until I manually change to 755 on CircleCI. The process crashes at 01-prepare. However, this behavior may not be the norm for local installations of terminus which explains why some users have passing tests.

Where would you suggest setting permissions at build? That would have to be executed around `terminus-build-tools-plugin/Commands/ProjectCreateCommand.php:285`, since the site is not available yet in `scripts/set-environment`, right?

greg-1-anderson:
There are a number of places where this could be done. Logically, I think I would do it in `BuildToolsBase::createFromSourceProject()`, right after the `composer create-project`. If you did it there, you'd have to confirm there was a scripts folder and then +x everything inside it.

Alternately, you could make a post-update command in the composer.json of this project. If you did it here, you could just +x a list of well-known scripts.

jraviotta:
Sounds good. I'll see if I can make that work. FYI i logged some other issues on  https://github.com/pantheon-systems/docker-build-tools-ci/issues that stem from running terminus build tools in a container rather than installed locally.

jraviotta:
Thanks for bearing with me. I'm a novice at collaborative development. I added the permission command in the post-update-cmd here. I am confused however, about how `BuildToolsBase` relates to `docker-build-tools-ci`. In trying to test edits, I searched in the container and located `terminus-plugins/terminus-build-tools-plugin/src/Commands/BuildToolsCommand.php` but there is no `createFromSourceProject` function so I'm not sure how I would be able to dry run edits.

greg-1-anderson:
The change to composer.json here look good to me. Remove the changes in .circleci and this is good to go.

jraviotta:
Reverted changes in .circleci. This should be good to go.

greg-1-anderson:
Thanks for your work here.

jraviotta:
Thanks Greg. Committing a milestone to my personal life repo: `-label="Developer v2.0"  -m="Committed an improvement to an open source project."` 🥇

PR #168, opened 2018-09-20
Try dev version of Drush 9 with the 9.x Behat Drush Endpoint.

greg-1-anderson:
This version works with Drush 9.4.0 installed on the platform. Note that we remove the site-local Drush so that it is possible to use either the global Drush 8 or the global Drush 9 on the platform. The Dashboard will automatically use the global Drush 8 when running the site audit command to generate the contents of the Status tab.


greg-1-anderson:
We still need to get to green at https://github.com/jhedstrom/drupalextension/pull/506, but we're almost there.

greg-1-anderson:
Closing in favor of #190

PR #178, opened 2019-01-02
composer run much faster on Drupal >8.5 projects

shaal:
https://www.jeffgeerling.com/blog/2018/make-composer-operations-drupal-way-faster-and-easier-on-ram


greg-1-anderson:
This was merged somewhere down the way. Thanks.

PR #187, opened 2019-03-12
Aligning versions in drush-ops/behat-drush-endpoint between composer.json and .lock

stevector:
A recent PR updated drush-ops/behat-drush-endpoint in composer.json but not composer.lock. That change means that nearly any `composer require` command will throw an error.


alexfornuto:
Closes https://github.com/pantheon-systems/documentation/issues/3890
Closes https://github.com/pantheon-systems/documentation/issues/4517

greg-1-anderson:
Whoops. Thanks for catching this. In some projects I run `composer validate` in the tests to catch this sort of error.

PR #195, opened 2019-03-23
BT-8: Gitlab support

rvtraveller:
Adds support for GitLab to the base upstream here, thereby enabling support via Terminus Build Tools. An example site demonstrating GitLab support can be found [here](https://gitlab.com/mindgrub_bthompson/bt-buildtools-d8-gl-5). This also adds support for code-style checking of custom modules in themes via the GitLab workflow.

Not added with these changes is pre-configured unit test support. Unit test support through PHPUnit on PHP 7.x requires a higher version of PHPUnit than shipped with Drupal, according to webflo/drupal-core-strict. This should change with future iterations of Drupal and can be added at a future date (at latest, when Drupal 8.8 comes out -- https://www.drupal.org/project/drupal/issues/3008870, https://www.drupal.org/node/2957906, and webflo/drupal-core-strict#7).


.ci/behat-test-cleanup.sh
```diff
@@ -0,0 +1,18 @@
...
+  exit 0;
+fi
+
+echo "::::::::::::::::::::::::::::::::::::::::::::::::"
+echo "Behat clean up on site: $TERMINUS_SITE.$TERMINUS_ENV"
+echo "::::::::::::::::::::::::::::::::::::::::::::::::"
+echo
+
+# Restore the backup made before testing
+terminus -n backup:restore $TERMINUS_SITE.$TERMINUS_ENV --element=database --yes
```
greg-1-anderson: I don't think we want to add `backup:create` and `backup:restore` to the Drupal tests by default.

.ci/behat-test.sh
```diff
@@ -0,0 +1,50 @@
...
+
+# Ping drush to start ssh with the app server
+terminus -n drush $TERMINUS_SITE.$TERMINUS_ENV -- version
+
+# Verbose mode and exit on errors
+set -ex
+
+# Start headless Chrome
+echo "\n Starting Chrome in headless mode ..."
+google-chrome-unstable --disable-gpu --headless --remote-debugging-address=0.0.0.0 --remote-debugging-port=9222 --no-sandbox </dev/null &>/dev/null &
```
greg-1-anderson: I don't think we use headless Chrome in any of the example tests. We should either add some js tests, or maybe just comment this out for folks who might want to start using it.

greg-1-anderson:
I think it's fine to skip the unit tests here for now, although in my experience, the older versions of phpunit still work on php 7+, they just emit a few warnings that do not interfere with test results. I have not tried running Drupal unit tests like this though

.ci/behat-test-cleanup.sh
rvtraveller: Took this out for now. Folks can add back in themselves if they want.

.ci/behat-test.sh
rvtraveller: I commented this out (vs removing it completely). Hopefully we'll be able to leverage it in the future and this way it is here easily for those folks who want to start using it.

rvtraveller:
The Drupal unit tests refuse to run under php 7 without a higher version of phpunit -- https://github.com/drupal/drupal/blob/646584b3f718897179589f4f014531e3d9365331/core/tests/bootstrap.php#L163. I think it is fine skipping it for now.

greg-1-anderson:
LGTM

PR #197, opened 2019-03-26
Run a visual regression test when composer.lock changes.

ataylorme:
Run a visual regression test when a pull request is created and either

* `composer.lock` IS modified and `composer.json` IS NOT modified or
* the string `visual regression test` exists in the latest commit message

When the visual regression test is run failures cause the build to fail by doing `exit 1`. If the visual regression test is not run or passes the build passes with `exit 0`.

`backstop.json.default` is the template used to build `backstop.json` with [jq](https://stedolan.github.io/jq/) dynamically inserting the proper URLs to test against.

Currently, `backstop.json.default` only tests the homepage at a variety of viewports. In a real project, the template should be updated to test multiple templates (single article, taxonomy, custom views, etc.).


pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1772/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
[CI job `1769`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1769) is proof the visual test won't run unless `composer.lock` has changed.

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1773/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
[CI job `1772`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1772) is proof the test fails if visual regression between `live` and the multidev fails.

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1774/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
And finally, [CI job `1774`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1774) is an example of everything passing when visual regression passes.

.circleci/config.yml
```diff
@@ -134,6 +134,25 @@ jobs:
                 name: remove transient test fixtures
                 command: ./.circleci/scripts/pantheon/09-cleanup-fixtures
 
+    visual_regression_test:
+        <<: *defaults
+        docker:
+            - image: ataylorme/docker-node-advanced-wordpress-on-pantheon:backstop-js
```
ataylorme: The only reason for a custom image here instead of using `backstopjs/backstopjs:latest` is the need for [`jq`](https://stedolan.github.io/jq/).

.circleci/set-up-globals.sh
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
```
ataylorme: A lot of this file is duplication/slight variation of [`/build-tools-ci/scripts/set-environment` in `docker-build-tools-ci`](https://github.com/pantheon-systems/docker-build-tools-ci/blob/4.x/scripts/set-environment) and [`.circleci/set-up-globals.sh` from `example-wordpress-composer`](https://github.com/pantheon-systems/example-wordpress-composer/blob/a0187e98cd5283dd580c775621553149929769bf/.circleci/set-up-globals.sh).

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1785/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1788/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1789/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1796/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
[build `1792`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1792) shows the visual regression test skipped when `composer.lock` has not changed.

[build `1796`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1796) shows the visual regression test is run when the string `visual regression test` is in the latest commit message.

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+else
+	# visual regression passed
+	REPORT_LINK="\n\nView the [visual regression test report]($DIFF_REPORT_URL)"
+	echo -e "\n\nVisual regression test passed!"
+	PR_MESSAGE="Visual regression test passed! $REPORT_LINK"
+fi
+
+# Post the image back to the pull request on GitHub
+echo -e "\nPosting visual regression results back to PR #$PR_NUMBER "
+curl -s -i -u "$CIRCLE_PROJECT_USERNAME:$GITHUB_TOKEN" -d "{\"body\": \"$PR_MESSAGE\"}" $GITHUB_API_URL/issues/$PR_NUMBER/comments > /dev/null
```
greg-1-anderson: If we added a `build:comment:add:pr` command, then this could instead be:
```
terminus build:comment:add:pr $PR_NUMBER --message="$message"
```
Using only existing commands, this could be:
```
terminus build:comment:add:commit --sha=$COMMIT_SHA --message="$message"
```
We'd have to look up $COMMIT_SHA via:
```
COMMIT_SHA=$(git rev-parse HEAD)
```

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+REPORT_LINK="[![Visual report]($DIFF_IMAGE_URL)]($DIFF_REPORT_URL)"
+
+if [[ ${VISUAL_REGRESSION_RESULTS} == *"Mismatch errors found"* ]]
+then
+	# visual regression failed
+	echo -e "\nVisual regression test failed!"
+	PR_MESSAGE="Visual regression test failed! $REPORT_LINK"
+else
+	# visual regression passed
+	REPORT_LINK="\n\nView the [visual regression test report]($DIFF_REPORT_URL)"
```
greg-1-anderson: Is there a reason why you use a different style report link for success than for failure?

.circleci/set-up-globals.sh
greg-1-anderson: See my comment: https://github.com/pantheon-systems/example-wordpress-composer/pull/85/files#r269383419

It would be better to adapt the `set-environment` script to be useful for multiple providers. We could pass in `CIRCLE_BUILD_NUMBER` et. al. as parameters, and it could set generic `TERMINUS_` and `CI_` variables for us.

This might necessitate a 5.x version of the build container, unless perhaps we make it backwards-compatible (read `CIRCLE_` variables if nothing passed in).

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+# Check if we are NOT on the master branch and this is a PR
+if [[ ${CURRENT_BRANCH} != "master" && -z ${CI_PR_URL} ]];
+then
+  echo -e "\nVisual regression tests will only run if we are not on the master branch and making a pull request"
+  exit 0;
+fi
+
+echo -e "\nProcessing pull request #$PR_NUMBER"
+
+LAST_GIT_COMMIT_MESSAGE=$(git log -1 --pretty=%B)
```
greg-1-anderson: It would be nifty if the code here could be written into a Terminus plugin, and called e.g. via `terminus visual-regression:check-needed`.  Default behavior could be as written, with options to control files needed / files not needed / special string in comment, etc.

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+        then
+            return 0;
+        fi
+    done <<< "$GIT_FILES_CHANGED"
+
+    return 1;
+}
+
+# Always run visual tests if "visual regression test" is in the last commit message
+if [[ ${LAST_GIT_COMMIT_MESSAGE} != *"visual regression test"* ]]
```
greg-1-anderson: I'd make the special string something like "[vr]". Existing example of a special string is "[ci-skip]" (recognized by most CI services).

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+
+# Ping the multidev environment to wake it from sleep
+echo -e "\nPinging the ${TERMINUS_ENV} multidev environment to wake it from sleep..."
+curl -I "$MULTIDEV_SITE_URL" >/dev/null
+
+# Ping the live environment to wake it from sleep
+echo -e "\nPinging the live environment to wake it from sleep..."
+curl -I "$LIVE_SITE_URL" >/dev/null
+
+# Check for custom backstop.json
```
greg-1-anderson: Maybe package this up as `terminus visual-regression:backstop:init`

greg-1-anderson: With this code in php rather than bash, `jq` becomes unnecessary

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+if [ ! -f backstop.json ]; then
+	# Create Backstop config file with dynamic URLs
+	echo -e "\nCreating backstop.js config file..."
+	cat backstop.json.default | jq ".scenarios[0].url = \"$LIVE_SITE_URL\" | .scenarios[0].referenceUrl = \"$MULTIDEV_SITE_URL\" " > backstop.json
+fi
+
+# Backstop visual regression
+echo -e "\nRunning backstop reference..."
```
greg-1-anderson: The rest of this could be `terminus visual-regression:backstop:run`. Pass the filepaths in as options.

.circleci/set-up-globals.sh
```diff
@@ -0,0 +1,76 @@
...
+	PR_BRANCH="pr-${PR_NUMBER}"
+	echo "export DEFAULT_ENV=pr-${PR_NUMBER}" >> $BASH_ENV
+	echo "export PR_NUMBER=${PR_NUMBER}" >> $BASH_ENV
+else
+	# otherwise make the branch name multidev friendly
+	if [[ $CURRENT_BRANCH == "master" ]]
+	then
+		echo "export DEFAULT_ENV=dev" >> $BASH_ENV
+	else
+		echo 'export DEFAULT_ENV=$(echo ${PR_ENV:-$CI_ENV} | tr '"'"'[:upper:]'"'"' '"'"'[:lower:]'"'"' | sed '"'"'s/[^0-9a-z-]//g'"'"' | cut -c -11 | sed '"'"'s/-$//'"'"')' >> $BASH_ENV
```
greg-1-anderson: Is this the old `pr-BRANCHNAME` code? I think we've removed that in the current script in the docker container.

greg-1-anderson: In any event, I think I'd rather enhance the script in the docker container than duplicate / replace it in the repository.

.ci/visual-regression-test.sh
rvtraveller: +1 to some form of this.

ataylorme: We could use `CIRCLE_SHA1`, which is _The SHA1 hash of the last commit of the current build._, though I like the idea of `build:comment:add:pr`

.ci/visual-regression-test.sh
ataylorme: An image of the report is only shown on failure so you can see what's going on from the PR comment without needing to view the report.

.ci/visual-regression-test.sh
ataylorme: I agree but that seems out of scope for an MVP. Also, since we are showing this off in the Drupalcon training I would rather be verbose with the logic so folks can see what is going on rather than hiding it behind a Terminus command.

.ci/visual-regression-test.sh
ataylorme: That is a good suggestion. Square brackets are special characters in Bash but I think they can just be escaped with a backslash.

.ci/visual-regression-test.sh
ataylorme: See [my comment above](https://github.com/pantheon-systems/example-drops-8-composer/pull/197#discussion_r269584572) about creating new Terminus commands.

We could perhaps remove the `jq` dependency and do the replacement with `sed` or something else already available.

.ci/visual-regression-test.sh
ataylorme: See [my comment above](https://github.com/pantheon-systems/example-drops-8-composer/pull/197#discussion_r269584572) about creating new Terminus commands.

ataylorme:
> In any event, I think I'd rather enhance the script in the docker container than duplicate / replace it in the repository.

That would be ideal for running the environment variable setup. However, if the `jq` requirement is removed I would still like to use the Node-based Backstop JS image `backstopjs/backstopjs:latest`.

> At a minimum, though, it would be good to write a short php script to init the backstop config file & commit it to the repo in the .ci directory. Then we could drop the jq requirement and use a standard container.

The image running this job does not have PHP. We could potentially run two jobs, Backstop set up and then Backstop itself but that does not seem ideal.

Overall, Terminus and Pantheon tools are needed for deployment to Pantheon and some of the other jobs.

However, I think it is important to show users how they can extend our examples without heavy reliance on Pantheon provided images or tools. The visual regression test is a great way to demonstrate how the CI process can be extended.

.ci/visual-regression-test.sh
greg-1-anderson: I'd rather make an extra `git` call than make this script more Circle-specific.

greg-1-anderson:
Yes, using `backstopjs/backstopjs:latest` would be much better than having our own custom image. I wasn't thinking about the fact that this image doesn't have php. Does it have Python or Ruby or any other language that might have some js-manipulating APIs?

greg-1-anderson:
I'd rather have this PR continue to use a custom container than replace `jq` with `sed`. If there isn't any way to manipulate the .json file e.g. with Python, then just switch to using the comment command in the Terminus Build Tools plugin, and we'll call this an MVP.

I disagree with the premise that the students need to see a page of bash instead of an encapsulated, reusable command, but we can put that off for a later improvement.

ataylorme:
> then just switch to using the comment command in the Terminus Build Tools plugin, and we'll call this an MVP.

If we are sticking with the Node Backstop image Terminus build tools is not available

ataylorme:
> If there isn't any way to manipulate the .json file e.g. with Python

We shouldn't introduce another server-side scripting language when we already have one (Node)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1812/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1813/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1813/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1816/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1817/artifacts/0/artifacts/backstop_data/html_report/index.html)

greg-1-anderson:
> We shouldn't introduce another server-side scripting language when we already have one (Node)

I just mentioned Python because it's often available in minimal containers. If that image has node, that will do. Thanks for rewriting that part.

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1817/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1817/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
> If that image has node, that will do. Thanks for rewriting that part.

It helps if I export the environment variables I wish to read in Node so that they are actually defined...

@greg-1-anderson this is ready for review again.

There is a new script `.ci/create-backstop-json-from-default.js` that copies `.ci/backstop.default.json` to `.ci/backstop.json`.

Without needing `jq` the image used for the visual regression job is now `backstopjs/backstopjs:latest`.

As an added benefit, the Node script now loops through all scenarios so that `.ci/backstop.default.json` can have more than the homepage if desired. Previously, just URLs of the first scenario were updated.

The string to search for in the commit messages is now `[vr]` as well.

I'm not sure how to avoid `.circleci/set-up-globals.sh` while using the `backstopjs/backstopjs:latest`.

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1820/artifacts/0/artifacts/backstop_data/html_report/index.html)

greg-1-anderson:
What if we saved $BASH_ENV in the cache in an earlier build step that used `docker-build-tools-ci`, and then restored it in the visual regression container?  We'd still need to make sure that environment variables appropriate to the CI service are set up for the artifacts directory and URL.

I'm working on making these scripts more generic. See https://github.com/pantheon-systems/docker-build-tools-ci/commit/48849f0b4f6f2ca6053dba322a9271e18c2ffad8 and https://github.com/pantheon-systems/example-drops-8-composer/pull/198.

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1842/artifacts/0/artifacts/backstop_data/html_report/index.html)

.circleci/config.yml
```diff
@@ -131,7 +131,6 @@ jobs:
             - run:
                 name: prepare database for site-under test
                 command: ./.circleci/scripts/pantheon/02-init-site-under-test-clone-existing
-                # command: ./.circleci/scripts/pantheon/02-init-site-under-test-reinstall-new
```
greg-1-anderson: This comment is supposed to stay here. Users can select between one of the two strategies. We could move the commit and / or change its contents to make this clearer if you prefer.

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1856/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1857/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1857/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1857/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1857/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1858/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1859/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1865/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
I'm not sure why the `BACKSTOP ERROR: Chromy error: Error. See scenario – Homepage (phone)` error on `backstop test` is happening.

I re-ran the job in SSH and all environment variables are defined properly and the generated `.ci/backstop.json` file is valid (output below).

```sh
cat .ci/backstop.json
{"viewports":[{"name":"phone","width":320,"height":480},{"name":"tablet_v","width":568,"height":1024},{"name":"tablet_h","width":1024,"height":768},{"name":"desktop","width":1920,"height":1080}],"scenarios":[{"label":"Homepage","url":"https://live-ci-example-d8-composer.pantheonsite.io/","referenceUrl":"https://pr-197-ci-example-d8-composer.pantheonsite.io/","hideSelectors":[],"selectors":["document"],"readyEvent":null,"delay":1500,"misMatchThreshold":0.1}],"paths":{"bitmaps_reference":"backstop_data/bitmaps_reference","bitmaps_test":"backstop_data/bitmaps_test","compare_data":"backstop_data/bitmaps_test/compare.json","casper_scripts":"backstop_data/casper_scripts"},"engine":"chrome","report":["CI"],"casperFlags":[],"debug":false,"port":3001}
```

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1869/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1885/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1895/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1896/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1896/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1896/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1896/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test failed! [Visual report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1896/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1900/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1904/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
I had to go all the way back to Backstop `3.9.5`, the last version where visual regression passed, before their flurry of releases in the last 24 hours.

ataylorme:
@greg-1-anderson we should probably merge [the Dockerfile PR](https://github.com/pantheon-systems/docker-build-tools-ci/pull/19 ) first then update the image declaration in this PR

.ci/create-backstop-json-from-default.js
```diff
@@ -0,0 +1,16 @@
+var fs = require('fs');
```
greg-1-anderson: It would be great to have some comments in this file explaining when it is called (and from where) and what it is for.

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,127 @@
...
+else
+	# visual regression passed
+	REPORT_LINK="\n\nView the [visual regression test report]($DIFF_REPORT_URL)"
+	echo -e "\n\nVisual regression test passed!"
+	PR_MESSAGE="Visual regression test passed! $REPORT_LINK"
+fi
+
+# Post the image back to the pull request on GitHub
+echo -e "\nPosting visual regression results back to PR #$PR_NUMBER "
+curl -s -i -u "$CI_PROJECT_USERNAME:$GITHUB_TOKEN" -d "{\"body\": \"$PR_MESSAGE\"}" $GITHUB_API_URL/issues/$PR_NUMBER/comments > /dev/null
```
greg-1-anderson: I guess you are using `curl` here rather than `terminus build:commit:comment:add` is that we're running in the backstop.js image, and Build Tools are not available.

.circleci/config.yml
```diff
@@ -38,9 +38,31 @@ defaults: &defaults
...
+            - image: quay.io/pantheon-public/build-tools-ci:visual-regression-vars
+
+        steps:
+            - checkout
+
+            - run:
+                # Set TERMINUS_ENV and related environment variables.
+                # https://github.com/pantheon-systems/docker-build-tools-ci/blob/4.x/scripts/set-environment
+                name: environment
+                command: /build-tools-ci/scripts/set-environment
```
greg-1-anderson: I'm still wary about running this script here. I think future enhancements are likely to cause frequent build failures here, as `set-environment` assumes that it is being run in `build-tools-ci` (since it is part of the build-tools-ci container). Running a script from a different container without using that container just does not seem like a good practice.

Let's try caching $BASH_ENV in an earlier stage and restoring + sourcing it here. Identify what other things from that script are necessary, and make a separate script to do just what is necessary here.

.ci/visual-regression-test.sh
ataylorme: Correct

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1937/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1940/artifacts/0/artifacts/backstop_data/html_report/index.html)

.circleci/config.yml
```diff
@@ -38,9 +38,35 @@ defaults: &defaults
 
 version: 2
 jobs:
+    configure_env_vars:
+        <<: *defaults
+
+        docker:
+            - image: quay.io/pantheon-public/build-tools-ci:visual-regression-vars
```
ataylorme: This needs to be `quay.io/pantheon-public/build-tools-ci:4.x` after https://github.com/pantheon-systems/docker-build-tools-ci/pull/19 is merged/deployed

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1945/artifacts/0/artifacts/backstop_data/html_report/index.html)

.ci/create-backstop-json-from-default.js
greg-1-anderson: Rather than comments that simply say what the next line is doing, I had in mind something more like:

> This file is called by visual-regression-test.sh whenever the user has not already provided
> a Backstop configuration file.  In this instance, a default configuration file will be written from
> a template.

.ci/visual-regression-test.sh
```diff
@@ -0,0 +1,132 @@
...
+# Ping the multidev environment to wake it from sleep
+echo -e "\nPinging the ${TERMINUS_ENV} multidev environment to wake it from sleep..."
+curl -I "$MULTIDEV_SITE_URL" >/dev/null
+
+# Ping the live environment to wake it from sleep
+echo -e "\nPinging the live environment to wake it from sleep..."
+curl -I "$LIVE_SITE_URL" >/dev/null
+
+# Check for custom backstop.json
+if [ ! -f ./.ci/backstop.json ]; then
```
greg-1-anderson: So, I just had a thought about this.  If there is no backstop.json file, then the script will be run, and it will 
fill in scenario.url with `process.env.MULTIDEV_SITE_URL`. This value changes with every PR, correct?

It seems to me that if there is no backstop.json, then the default config file should be copied over to backstop.json, but thereafter, the contents of backstop.json should be replaced so that the dynamic values can be updated regardless of whether the user has provided customized values.

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1951/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1955/artifacts/0/artifacts/backstop_data/html_report/index.html)

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/1960/artifacts/0/artifacts/backstop_data/html_report/index.html)

PR #201, opened 2019-04-05
BT-25: Composer Lock Updater for GitHub/CircleCI

ataylorme:
# BT-25: Composer Lock Updater for GitHub/CircleCI

[Composer Lock Updater](https://github.com/danielbachhuber/composer-lock-updater) for GitHub/CircleCI

**Done when:**

Automatic PR on GitHub for Circle only

**Out of scope:**

Other providers (GitLab / BitBucket)


.ci/composer-lock-updater.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# Exit on errors
+set -e
+
+# Install CLU
+echo "export PATH='$HOME/.composer/vendor/bin:$PATH'" >> $BASH_ENV
+source $BASH_ENV
+composer global require danielbachhuber/composer-lock-updater
+
+# Install HUB
```
greg-1-anderson: Hub is already installed in our container

.ci/composer-lock-updater.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+# Exit on errors
+set -e
+
+# Install CLU
+echo "export PATH='$HOME/.composer/vendor/bin:$PATH'" >> $BASH_ENV
+source $BASH_ENV
+composer global require danielbachhuber/composer-lock-updater
```
greg-1-anderson: Let's install this in the container, not in each sample on every test run.

.ci/composer-lock-updater.sh
```diff
@@ -0,0 +1,20 @@
...
+# Install HUB
+cd $HOME
+wget -O hub.tgz https://github.com/github/hub/releases/download/v2.2.9/hub-linux-amd64-2.2.9.tgz
+tar -zxvf hub.tgz
+echo "export PATH='$PATH:$PWD/hub-linux-amd64-2.2.9/bin/'" >> $BASH_ENV
+source $BASH_ENV
+
+# Run composer lock updater
+cd $CIRCLE_WORKING_DIRECTORY
+clu https://${GITHUB_TOKEN}:x-oauth-basic@github.com/${CI_PROJECT_USERNAME}/${CI_PROJECT_REPONAME}.git
```
greg-1-anderson: I think that @rvtraveller 's most recent PR adds the project URL with the provider-specific authentication token to the build metadata file. We should use that.

.circleci/config.yml
```diff
@@ -176,14 +189,26 @@ workflows:
...
+        - visual_regression_test:
             requires:
                 - configure_env_vars
                 - functional
+        # Composer lock updater
+        - composer_lock_updater
+#  auto_update_dependencies:
+#    triggers:
+#        - schedule:
+#            cron: "5 4 * * *"
```
greg-1-anderson: This is five minutes after 4am UTC every day? Since Circle does not support anything beyond the most basic cron schedule syntax, our choice is to run daily or monthly.

greg-1-anderson:
Today, Build Tools instantiated the repository provider via:
```
$this->inferGitProviderFromUrl($remoteUrl);
```
In most cases, the remote URL (to the repository service) is recovered via `retrieveRemoteUrlFromBuildMetadata()`. If you look at `build:env:delete`, though, it also uses:
```
$remoteUrlFromGit = exec('git config --get remote.origin.url');
```
When running on the CI provider, the current git repository should always have the remote.origin.url pointed at the Repository provider.  We could use that technique to look up the URL and then use with `inferGitProviderFromUrl()` to get the repository provider.

The repository provider should be enhanced (in the GitProvider interface) to add a new method that will add the authentication token to an unauthenticated URL.

.circleci/config.yml
ataylorme: > This is five minutes after 4am UTC every day?

Yes.

> Since Circle does not support anything beyond the most basic cron schedule syntax, our choice is to run daily or monthly.

That is not true. This is pulled from [`pantheon-systems/example-terminus-auto-update-script`](https://github.com/pantheon-systems/example-terminus-auto-update-script) where it does, indeed, run at `4:05` daily.

See [this CircleCI docs](https://circleci.com/docs/2.0/workflows/#specifying-a-valid-schedule). Values for all 5 places are supported but Cron step syntax is not.

Anyway, I updated it to hourly. Once this PR passes I will remove the job from the workflow that runs on commit and uncomment the cron workflow.

ataylorme:
Improving build tools and perhaps even adding a command like `build:composer:update-pr` to run Composer Lock Updater seems ideal.

However, it falls out of scope for `BT-25`

```
Done when:

Automatic PR on GitHub for Circle only

Out of scope:

Other providers (GitLab / BitBucket)
```

.circleci/config.yml
greg-1-anderson: You're right, hourly is also an option, but that's too frequent. I thought daily was too frequent, and was hoping for weekly, but alas, that is not supported. Given the option between every minute, hourly, daily and monthly (or yearly), I think we should do daily.

We could have our script no-op one time out of seven, but I don't want to add the complexity.

ataylorme:
@greg-1-anderson `find .circleci/scripts/pantheon/ -type f | xargs chmod 755` is failing. Any reason that or `"find tests/scripts/ -type f | xargs chmod 755"` are necessary anymore?

It is in the [`post-update-cmd` section of `composer.json`](https://github.com/pantheon-systems/example-drops-8-composer/blob/master/composer.json#L73:L77)

Since it is in `post-update-cmd` it doesn't happen on `composer install` so I think we've missed it to this point.

greg-1-anderson:
The scripts moved to `.ci/scripts`. However, I think we can just remove this.

.circleci/config.yml
```diff
@@ -151,13 +151,17 @@ jobs:
...
+        docker:
+            - image: quay.io/pantheon-public/build-tools-ci:4.x
         steps:
             - checkout
 
             - run:
                 name: setup-environment-vars
                 command: /build-tools-ci/scripts/set-environment
 
+            - run: source $BASH_ENV
```
greg-1-anderson: Every CircleCI phase runs in its own shell, so I don't think that the `source $BASH_ENV` here has any effect on later phases.

Also, I believe that CircleCI automatically runs `source $BASH_ENV` at the beginning of each `run` line.

ataylorme:
[A PR from `clu`](https://github.com/pantheon-systems/example-drops-8-composer/pull/204)!

ataylorme:
@greg-1-anderson suggested https://github.com/g1a/composer-test-scenarios/blob/master/src/CommandProvider.php#L24 for the long-term as [CLU](https://github.com/danielbachhuber/composer-lock-updater) is hard-coded to GitHub and requires [hub](https://github.com/github/hub)

greg-1-anderson:
That command updates composer.lock without installing `vendor`. I'm pretty sure it respects platform.php etc.

greg-1-anderson:
A [passing test created by composer-lock-updater!](https://github.com/pantheon-systems/example-drops-8-composer/pull/209)

PR #250, opened 2019-06-06
Add exception for services.yml

cmcky:
services.yml should not be ignored and is causing launch check warnings on Pantheon dashboard.


greg-1-anderson:
Yes, the existing entry is overly broad. This is one way to fix it. Thanks.

PR #255, opened 2019-07-16
Reorganize the CI scripts and directories to be more readable

### LINKED ISSUE #254: Reorganize the CI scripts and directories to be more readable
Parity work from [https://github.com/pantheon-systems/example-wordpress-composer/pull/136](https://github.com/pantheon-systems/example-wordpress-composer/pull/136)

ataylorme:
Closes #254. See https://github.com/pantheon-systems/example-wordpress-composer/pull/136


ataylorme:
Behat tests are failing and I'm not sure why maybe a misconfiguration with the aliases. The `.feature` file hasn't changed. @stevector any thoughts here?

.ci/deploy/pantheon/dev-multidev
```diff
@@ -0,0 +1,42 @@
...
+# Otherwise a multidev environment is used.
+#
+
+# Authenticate with Terminus
+terminus -n auth:login --machine-token="$TERMINUS_TOKEN"
+
+if [[ $CI_BRANCH != "master" ]]
+then
+    # Create a new multidev environment (or push to an existing one)
+    terminus -n build:env:create "$TERMINUS_SITE.dev" "$TERMINUS_ENV" --yes
```
davereid: This should have --clone-content also?

ataylorme: We can give it a shot but there is [a prior job that runs `multidev:create`](https://github.com/pantheon-systems/example-drops-8-composer/pull/255/files#diff-2bb1205325a28694f90c34db9aa1b73fR15) so the multidev environment is already created by the time we get to this step. During multidev creation content is cloned from the source environment (`dev`).

I used `build:env:create` just in case, as it will create the multidev if it does not already exist but here acts more like `build:env:push`.

I'm not sure if the addition of `--clone-content` means that content cloning will happen on each CI build. If so, it probably isn't desirable as all previous changes to that multidev, such as configuration, will be lost when the database is overwritten during the clone.

ataylorme:
@stevector thanks for dropping aliases for using Terminus as an alternate Drush binary. There seems to be an issue though as the Drush commands on Pantheon being run are `drush behat ...` instead of `drush`

stevector:
@greg-1-anderson, do you know what's going on here? I thought it might be an incompatibility between drush and the behat endpoint. https://github.com/pantheon-systems/example-drops-8-composer/pull/255/commits/02528361c45584674f75c31071e08bdc93e7e165

But that commit didn't change the error message.

ataylorme:
@greg-1-anderson could this be an issue in `drupal-driver`? https://github.com/jhedstrom/DrupalDriver/pull/91

ataylorme:
https://github.com/jhedstrom/DrupalDriver/issues/90 as well

ataylorme:
@dsnopek seems you ran into the issue we are having with Behat in the past. Happen to have any suggestions for [our failing Behat job](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/2782)

ataylorme:
@greg-1-anderson I added a simple scenario that visits the home page (below) and it works. The issues seem to be specific to the Drush driver. I've tried but can't get it going and am beyond my limitations trying to debug it. Any help is appreciated.

```
Scenario: Verify the homepage
    When I am on the homepage
    Then I should be on the homepage
    And the response status code should be 200
```

ataylorme:
I went back to using Selenium 2 and not the Chrome Driver and [Behat still fails](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/2992). I think it may be an admin login issue as `Create many terms` fails with a `403` and lands on the URL https://pr-255-ci-example-d8-composer.pantheonsite.io/user/logout

ataylorme:
@stevector and @greg-1-anderson what do you think of updating the feature file to remove the scenarios that work with users, nodes and taxonomies?

@populist and I were talking yesterday and while this has worked in the past according to [the Drupal Driver docs](https://drupal-drivers.readthedocs.io/en/1.1/drivers.html) the Drush driver does *not* support creating users, nodes or taxonomy terms.

I've tried different combinations of the following without success:

- Adding Drush alias
- Using Chrome
- Using Selenium
- Updating `drupal/drupal-driver` and `drupal/drupal-extension` to the new, RC versions

Alternatively, if anyone can help solve the Drush Driver + users, nodes and taxonomy issues please do. I've spent too long on this as-is.

ataylorme:
Hmm, the [docs for the latest RC version](https://behat-drupal-extension.readthedocs.io/en/master/drivers.html) say that node and term creation are supported with the Drush driver.

ataylorme:
Yay, [a passing build](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3046) from commit `b0f3974b6afd1ec0d760835b9c54d9655fe20a30`.

ataylorme:
A [failing build](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3052) with:

- `drupal-extension` v4 RC
- `drupal/drupal-driver` v2 RC
- Headless Chrome
- Drush Driver
- node creation and taxonomy term steps failing
- User creations steps working

ataylorme:
Another [passing build](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3058) from `f91c9f69fbda702ae78d7034799e6242e8b9f930` with:

- `drupal-extension` v4 RC
- `drupal/drupal-driver` v2 RC
- Headless Chrome
- Drush Driver
- **No** node creation or taxonomy term steps
- User creations steps working

Failures correspond with the lack of node creation and taxonomy support in the Drush driver.

ataylorme:
@greg-1-anderson perhaps `drush-ops/behat-drush-endpoint` isn't working as intended here?

greg-1-anderson:
At one time, the Drush driver only supported creating users, not nodes or taxonomy terms, as provided by the Drupal driver. However, I contributed code that added taxonomy term and node support to the Drush driver.

Is there some reason you upgraded to `"drupal/drupal-extension": "v4.0.0rc1@dev",`? Try again with 3.x and see if it works. It could be a regression in the Drupal extension.

ataylorme:
@greg-1-anderson I've tried both `3.x` and `4.x` of `drupal/drupal-extension`. I've also tried Drush 8 and Drush 9 and saw no difference.

A [failing build](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3076) from commit `1dba2910b9bda30610a5d83bb773386adf5085c3` with:

- `drupal/drupal-driver` `v1.4.0`
- `drupal/drupal-extension` `v3.4.1`
- `drush-ops/behat-drush-endpoint` `9.3.0`
- Headless Chrome
- Drush Driver
- node creation and taxonomy term steps failing
- User creations steps working

> However, I contributed code that added taxonomy term and node support to the Drush driver.

I've identified taxonomy term and node support as the difference between passing tests and failing ones.

> It could be a regression in the Drupal extension.

If so, can we:
1) Open an issue on that repository
1) Remove use of terms and nodes in the Behat tests here until it is resolved

The purpose of this PR is to make the CI script paths/structure more readable. The context of our demo Behat feature files seems less relevant.

ataylorme:
The [latest passing build](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3082) `b12827157eac37fde3ba572bbbe97d1d9004fbf9`, which I want to merge, has:

- `drupal/drupal-driver` `v1.4.0`
- `drupal/drupal-extension` `v3.4.1`
- `drush-ops/behat-drush-endpoint` `9.3.0`
- Headless Chrome
- Drush Driver
- **No** node creation and taxonomy term steps
- User creations steps
- Drush `9`
- PHP `7.3`

ataylorme:
@greg-1-anderson help spotting the difference between what is failing here and [the passing build on `master`](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/2906) would be appreciated.

I've compared the `composer.lock` file and the `drupal/drupal-driver`, `drupal/drupal-extension`, and `drush-ops/behat-drush-endpoint` are all the same.

The Behat commands are the same, just ported to re-organized files.

The only other change is using Headless Chrome but the steps that don't use the Drush API driver, like the homepage `200` response test, pass. I've also tried rolling back headless Chrome and using Selenium without success.

greg-1-anderson:
First thing I checked is whether or not the `behat` command existed in your test site:
```
$ drush @ci-example-d8-composer.pr-255 help behat
Invalid command behat.                                           
```
Nope, not there.

Next question is, is it really there, and Drush just can't find it, or is it completely gone? I cloned the test site locally and took a look.
```
$ ls drush
reinstall.drush.inc
```
Hm, it is really really gone.

Control: (on the 'master' branch of the same site)
```
$ ls drush
Commands		reinstall.drush.inc
$ ls drush/Commands/contrib/behat-drush-endpoint/
BehatDrushEndpointCommands.php	behat.inc			composer.lock
README.md			composer.json
```
That's what it should look like. Why does your test site not have that? As you already observed, your composer.json and composer.lock indicate that the Behat Drush Endpoint should be present.

The answer is in your .gitignore file.
```
$ git diff master .gitignore
diff --git a/.gitignore b/.gitignore
index b228661c4..cd80bd675 100644
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,48 @@
+###
+### GitHub repository .gitignore section
+###
+
+# Ignore directories generated by Composer
+/drush/contrib/
+/drush/Commands/
+/vendor/
+/web/core/
+/web/modules/contrib/
+/web/themes/contrib/
+/web/profiles/contrib/
+/web/private/scripts/quicksilver
+/console/
+
+# Ignore scaffold files
+web/.csslintrc
+web/.editorconfig
+web/.eslintignore
+web/.eslintrc
+web/.gitattributes
+web/.htaccess
+web/.ht.router.php
+web/autoload.php
+web/index.php
+web/robots.txt
+web/sites/default/default.services.pantheon.preproduction.yml
+web/sites/default/default.services.yml
+web/sites/default/default.settings.php
+web/sites/default/settings.pantheon.php
+web/sites/development.services.yml
+web/sites/example.settings.local.php
+web/sites/example.sites.php
+web/update.php
+web/web.config
+
+# Add directories containing build assets below.
+# Keep all additions above the "cut" line.
+
+# This distinction is only important when using this
+# repository as a custom upstream. The .gitignore file
+# is not modified in the GitHub PR workflow.
+
 
+# :::::::::::::::::::::: cut ::::::::::::::::::::::
 
 ###
 ### Pantheon site .gitignore section
```

It looks as if for some reason, your test site is not correctly running the prepare-for-pantheon step that trims the .gitignore to the state it should be in on Pantheon. Given this defect, I am sort of surprised that your site works as well as it does; I would expect other things to also be wrong.

I leave it to you to investigate why the build step has gone awry.

ataylorme:
@greg-1-anderson thank you for your help diagnosing the issue of the `drush` directory not being deployed to Pantheon.

Neither the `.gitignore` file or the `drush` directory were being persisted from the PHP build job to the deploy job.

I did two things to fix this:
1) Persist the `drush` directory from the PHP build job to the deploy job.
	- Do you think we should persist and [`rsync`](https://github.com/pantheon-systems/example-drops-8-composer/blob/reorganize-ci-scripts/.circleci/config.yml#L138:L156) the entire docroot instead of [specific directories](https://github.com/pantheon-systems/example-drops-8-composer/blob/reorganize-ci-scripts/.circleci/config.yml#L116:L121)?
1) Do `composer run prepare-for-pantheon` just before deploying to Pantheon with `build:env:create`. The reason I did this instead of persisting `.gitignore` is it allows other build jobs, aside from PHP, such as Node, to run and `prepare-for-pantheon` will remove and `.git` directories, etc. from all build jobs.

[Here is a passing build with node and taxonomy terms steps](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3163) 🎉

.ci/deploy/pantheon/dev-multidev
```diff
@@ -0,0 +1,45 @@
...
+# If exported configuration is available, then import it.
+if [ -f "config/system.site.yml" ] ; then
+  terminus -n drush "$TERMINUS_SITE.$TERMINUS_ENV" -- config-import --yes
+fi
+
+# Clear Drupal cache
+terminus -n drush "$TERMINUS_SITE.$TERMINUS_ENV" -- cr
+
+# Clear the environment cache
+terminus -n env:clear-cache $TERMINUS_SITE.$TERMINUS_ENV
```
greg-1-anderson: Doesn't this call `drush cr` as part of its workflow?

.ci/test/behat/cleanup
```diff
@@ -0,0 +1,38 @@
...
+
+#
+# This script runs cleanup after Behat tests complete.
+#
+
+# Authenticate with Terminus
+terminus -n auth:login --machine-token="$TERMINUS_TOKEN"
+
+# Restore the DB backup made before testing
+terminus -n backup:restore $TERMINUS_SITE.$TERMINUS_ENV --element=database --yes
```
greg-1-anderson: I'm still -1 on the backup / restore workflow for tests. I think it's dangerous; far better to make non-destructive tests. I know you typically use this pattern on the WordPress side, but I thought that was mostly due to the fact that  you didn't have the features of the Drupal Driver to create fixtures on the fly that are automatically deleted at the end of a test.

If we do employ a backup/restore workflow, perhaps it could be disabled by default, and folks could set a well-known environment variable in their CI environment variables to turn it on.

.ci/test/behat/cleanup
```diff
@@ -0,0 +1,38 @@
...
+# Store artifacts
+export ARTIFACTS_DIR_URL="${CI_BUILD_URL}/artifacts/${CI_NODE_INDEX}/artifacts"
+
+# Create the artifacts directory if needed
+if [ ! -d "$ARTIFACTS_FULL_DIR" ]; then
+  mkdir -p $ARTIFACTS_FULL_DIR
+fi
+
+
+# Copy behat-screenshots files to ARTIFACTS_FULL_DIR
```
greg-1-anderson: This is awesome.

.ci/test/static/run
```diff
@@ -0,0 +1,28 @@
...
+#
+
+# Run composer install to get the vendor directory
+composer install --no-ansi --no-interaction --optimize-autoloader --no-progress
+
+# Delete and reset the web directory
+# This is necessary so that third-party
+# dependencies are not tested
+rm -rf web
+git checkout -- web
```
greg-1-anderson: Instead of running `composer install` and then cleaning the web directory, could we just add the linter and code-sniff binaries to our docker container?

I guess this way allows the client to be in control of the version of the linter used (e.g. if using a newer version of PHP, or if there is a bugfix in the linter). Maybe that's rare?

If you prefer to stick with this technique, I would recommend replacing the rm / checkout with:
```
git clean -ffdX web
```

.ci/test/static/run
```diff
@@ -0,0 +1,28 @@
...
+
+# Lint php code for syntax errors
+composer -n lint
+
+# Check coding standards
+composer -n code-sniff
+
+# Run unit tests
+composer -n unit-test
```
greg-1-anderson: The unit test might fail after cleaning the `web` directory.  We could lint / sniff without running `composer install`, but running the unit tests will likely need vendored code, so I suppose there's little point in avoiding the install.

The efficient way to resolve this would be to re-order the tests so that the unit tests ran before cleaning the web directory. That's probably better than running `composer install` twice, even though we might otherwise prefer to lint and sniff first.

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+      - restore_cache:
+          keys:
+            - composer-cache-{{ checksum "composer.lock" }}
+            - composer-cache-
+
+      - run:
+          # Set TERMINUS_ENV and related environment variables.
+          # https://github.com/pantheon-systems/docker-build-tools-ci/blob/6.x/scripts/set-environment
+          name: setup-environment-vars
+          command: /build-tools-ci/scripts/set-environment
```
greg-1-anderson: Since we're bothering to save $BASH_ENV, maybe we should make this step depend on configure_env_vars so that we can be consistent about including it here?

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+            - composer-cache-{{ checksum "composer.lock" }}
+            - composer-cache-
+            - vendor-{{ checksum "composer.lock" }}
+            - vendor-
+
+      - run:
+          # Set TERMINUS_ENV and related environment variables.
+          # https://github.com/pantheon-systems/docker-build-tools-ci/blob/6.x/scripts/set-environment
+          name: setup-environment-vars
+          command: /build-tools-ci/scripts/set-environment
```
greg-1-anderson: Here too.

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+          key: vendor-{{ checksum "composer.lock" }}
+          paths:
+            - $CIRCLE_WORKING_DIRECTORY/vendor
+
+      - persist_to_workspace:
+          root: .
+          paths:
+            - web
+            - vendor
+            - drush
```
greg-1-anderson: How do we know which paths need to be persisted?  e.g. what if the user's build step writes assets to another location?

Can we persist dynamic paths?

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+      - run:
+          # Set TERMINUS_ENV and related environment variables.
+          # https://github.com/pantheon-systems/docker-build-tools-ci/blob/6.x/scripts/set-environment
+          name: setup-environment-vars
+          command: /build-tools-ci/scripts/set-environment
+
+      # Rsync the built web directory
+      - run:
+          name: sync web
+          command: rsync -az /tmp/web/ web
```
greg-1-anderson: Should be able to do this all in one rsync. Presuming that there is nothing in the workspace except what you want to restore, then `rsync -az /tmp/ .` should do the trick. I guess you have persisted the bash env; you could simply ignore that with an `--exclude`.

.circleci/config.yml
greg-1-anderson: Also, this is persisting files that are committed to the repository. That's inefficient, although I don't know that it's a problem.

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+    <<: *defaults
+    docker:
+      - image: backstopjs/backstopjs:4.1.9
+    steps:
+      - checkout
+
+      - attach_workspace:
+          at: /tmp/workspace
+
+      - run: cp /tmp/workspace/bash_env.txt $BASH_ENV
```
greg-1-anderson: I guess the reason this is the only step where bash_env is restored is that this container cannot run the set-environment script.

.circleci/config.yml
```diff
@@ -38,177 +38,232 @@ defaults: &defaults
...
+    <<: *defaults
+
+    steps:
+      - checkout
+
+      - run:
+          # Set TERMINUS_ENV and related environment variables.
+          # https://github.com/pantheon-systems/docker-build-tools-ci/blob/6.x/scripts/set-environment
+          name: setup-environment-vars
+          command: /build-tools-ci/scripts/set-environment
```
greg-1-anderson: I recall that the `set-environment` script calls `terminus auth:login` as a side-effect. I am now regretting that decision, as not every Circle CI step that calls set-environment necessarily needs to log in. Also, if we save/restore bash_env, then steps that need the terminus login won't get it.

We also made it the responsibility of the set_env script to re-install the Terminus Build Tools plugin if requested. I think that only the Terminus Build Tools plugin project itself needs that.

We'd need to make a 7.x version of the docker container to fix those problems, though.

ataylorme: > We'd need to make a 7.x version of the docker container to fix those problems, though.

Agreed, I've opened issues [`32`](https://github.com/pantheon-systems/docker-build-tools-ci/issues/32) and [`33`](https://github.com/pantheon-systems/docker-build-tools-ci/issues/33) on that repository

.circleci/config.yml
ataylorme: That is correct

.circleci/config.yml
ataylorme: That's what I was asking with

> Do you think we should persist and rsync the entire docroot instead of specific directories?

.circleci/config.yml
ataylorme: > How do we know which paths need to be persisted? e.g. what if the user's build step writes assets to another location?

Remember that this repository is an example, meant to be hacked/tweaked. If a user modifies the build step, adds another build step, etc. they should also adjust what is being persisted between jobs to fit their needs.

.circleci/config.yml
ataylorme: > Since we're bothering to save $BASH_ENV, maybe we should make this step depend on configure_env_vars so that we can be consistent about including it here?

Considering `/build-tools-ci/scripts/set-environment` takes less than ` second to run I don't think there is anything to be gained here.

.ci/test/static/run
ataylorme: Updated in `54ad541439ac6d0f8886a9c51efe4c3d119bcb66`

.ci/test/static/run
ataylorme: Updated in `c2aeff7e22c7a35690a661dbce1d5f1ac33487b3`. I used `git clean -fdX web` as I think the double `f` was a typo.

.ci/test/behat/cleanup
ataylorme: I've had issues reconfiguring the directory [Behat fail aid](https://packagist.org/packages/genesis/behat-fail-aid) saves screenshots to but I was successful in [writing my own screenshot step](https://github.com/ataylorme/ataylorme-wordpress/blob/master/tests/behat/helper_classes/Contexts/MiscContext.php#L38:L58) using the Chrome Driver that works well. Think that is worth adding here?

I could use it in a few places in the existing example feature file, e.g. when created articles, users or terms are viewed in the admin.

.ci/test/behat/cleanup
ataylorme: > I know you typically use this pattern on the WordPress side, but I thought that was mostly due to the fact that you didn't have the features of the Drupal Driver to create fixtures on the fly that are automatically deleted at the end of a test.

That is true. Feel free to push a commit to this PR that does the cleanup with the features of the Drupal Driver.

.ci/deploy/pantheon/dev-multidev
ataylorme: I'm not sure. The [Terminus source just shows the clear cache workflow being called](https://github.com/pantheon-systems/terminus/blob/fcbfecb8b74926e9539f2eca87643618d6256338/src/Models/Environment.php#L144).

If you're confident the clear cache workflow clears Drush cache, and not just Varnish, we can remove the `drush cr` call.

However, I don't think to leave it in hurts in case the workflow behavior ever changes.

.ci/test/static/run
ataylorme: This should probably be ported to `example-wordpress-composer` as well

.ci/test/static/run
ataylorme: This should probably be ported to `example-wordpress-composer` as well

.circleci/config.yml
ataylorme: @greg-1-anderson after reading [the CircleCI doc on `persist_to_workspace`](https://circleci.com/docs/2.0/configuration-reference/#persist_to_workspace) and seeing

> If multiple concurrent jobs persist the same filename then attaching the workspace will error.

I don't think we should persist the whole docroot. While it will be fine for a single build job, if multiple build jobs persist the entire docroot the build will fail as they will persist files with the same name.

The way it is set up currently, each build job only persists the files it builds. While this is a bit more work to maintain the list of files to persist for each build job it is the only way to avoid a conflict.

.circleci/config.yml
ataylorme: An alternative could be to persist the `/build-tools-ci/scripts/set-environment`, rather than the resulting `$BASH_ENV` file, and run the `set-environment` script directly.

pantheon-upstream:
Visual regression test passed! 

View the [visual regression test report](https://circleci.com/gh/pantheon-systems/example-drops-8-composer/3185/artifacts/0/artifacts/backstop_data/html_report/index.html)

ataylorme:
The link to the multidev being sent to the PR on GitHub from `terminus build:env:create` is missing.

`build:env:create` calls `build:comment:add:commit`, which calls `$this->git_provider->commentOnCommit`.

Perhaps we need to open an issue on the Build Tools repository. I think instead of commenting on the commit the multidev link should be added to the pull request.

PR #257, opened 2019-07-18
Skip running build:env:create when the source and target are both dev

davereid:
This ends up destroying the files directory on the dev environment due to https://github.com/pantheon-systems/terminus/issues/1990


davereid:
Hold on, not calling build:env:create when this is run on dev means that we end up not getting any code updates deployed to dev. I would have figured that the 05-merge-master does that, but I guess it doesn't?

davereid:
Ah so build:env:merge is only helpful if it's run on the PR/CI environment, not on dev environment since it has an early return if run on dev.

PR #261, opened 2019-08-14
Add console directory to .gitignore

aaronbauman:
during composer install, drupal/console-en puts its language files in root at /console, which should be added to the ignore list.


greg-1-anderson:
LGTM. Thanks.

PR #264, opened 2019-08-19
Update composer.json drush version and update composer.lock file.

szipfel:
When following the build tool instructions "Start a Pull Request" part here: https://pantheon.io/docs/guides/build-tools/new-pr the build fails due to `Drush version (8.2.3) is incompatible with Pantheon aliases` error. 

Screen Shot 2019-08-16 at 2 21 46 PM

This PR has an update to 8.3 in the composer.json and the updated composer.lock file.


greg-1-anderson:
Thanks for this update.

PR #275, opened 2019-09-09
Update bb pipeline config for refactored ci scripts

aaronbauman:
I've done some preliminary tests with this bitbucket pipelines config, and builds are passing.

There's a bug in bitbucket: BITBUCKET_PR_ID is not set correctly

Otherwise seems to be working well


ataylorme:
@aaronbauman I'm going to work on getting [advanced configuration](https://confluence.atlassian.com/bitbucket/configure-bitbucket-pipelines-yml-792298910.html#Configurebitbucket-pipelines.yml-Advancedconfiguration) with multiple steps and parallel jobs going rather than one, monolithic repository.

See [`bitbucket-pipelines.yml` in the `bitbucket-pipelines` branch of `example-wordpress-composer`](https://github.com/pantheon-systems/example-wordpress-composer/blob/bitbucket-pipelines/bitbucket-pipelines.yml) for what I have so far.

I need to test it out. Once I get things working I'll open a PR on `example-wordpress-composer` and then mirror it here.

aaronbauman:
Config definitely leaves a lot to be desired, but at least it builds successfully.

Another problem: it runs on every commit.
Unlike Circle, looks like changing this is a config-level setting, rather than a project setting.

ataylorme:
@aaronbauman sorry, I linked the wrong branch before. I am working in the [`pipelines-config` branch of `example-wordpress-composer`](https://github.com/pantheon-systems/example-wordpress-composer/tree/pipelines-config).

In [the `bitbucket-pipelines.yml` file there](https://github.com/pantheon-systems/example-wordpress-composer/blob/pipelines-config/bitbucket-pipelines.yml) I am using `pull-requests` pipeline filtered to `**` (all PRs) and the `branches` pipeline (filtered to `master`) instead of the `default` pipeline (which builds all commits).

ataylorme:
@aaronbauman an initial PR is up at https://github.com/pantheon-systems/example-wordpress-composer/pull/171

It is a bit different than what is done here. I'd love to have your thoughts on that approach

aaronbauman:
cool, i will try it out

bitbucket-pipelines.yml
```diff
@@ -1,17 +1,122 @@
...
+      script:
+        - *bash_env_export
+        - *bash_env_source
+        - ./.ci/deploy/pantheon/dev-multidev
+      artifacts:
+        - web/**
+        - vendor/**
+        - drush/**
+        - console/**
+  - parallel:
```
aaronbauman: Any reason not to run Behat and Visual Regression in parallel @ataylorme ?

bitbucket-pipelines.yml
```diff
@@ -1,17 +1,122 @@
...
+      script:
+        - *bash_env_export
+        - export CI_BRANCH=${BITBUCKET_BRANCH:-master} && echo $CI_BRANCH
+        - export PR_NUMBER=$BITBUCKET_PR_ID && echo $PR_NUMBER
+        - export CI_PR_URL=https://bitbucket.org/$BITBUCKET_REPO_OWNER/$BITBUCKET_REPO_SLUG/pull-requests/$BITBUCKET_PR_ID && echo $CI_PR_URL
+        - export CI_BUILD_NUMBER=$BITBUCKET_BUILD_NUMBER && echo $CI_BUILD_NUMBER
+        - export CI_PROJECT_USERNAME=$BITBUCKET_REPO_OWNER && echo $CI_PROJECT_USERNAME
+        - export CI_PROJECT_REPONAME=$BITBUCKET_REPO_SLUG && echo $CI_PROJECT_REPONAME
+        - export CI_PROJECT_NAME=$BITBUCKET_REPO_FULL_NAME && echo $CI_PROJECT_NAME
+        - ./.ci/build/set-environment
```
aaronbauman: using local set-environment script here, so that we can run build tools dev-master on top of arbitrary docker image

bitbucket-pipelines.yml
```diff
@@ -1,17 +1,122 @@
...
+            - ./.ci/test/behat/initialize
+            # Start headless Chrome
+            - google-chrome --disable-gpu --headless --remote-debugging-address=0.0.0.0 --remote-debugging-port=9222 --no-sandbox </dev/null &>/dev/null &
+            - ./.ci/test/behat/run
+          after-script:
+            - ./.ci/test/behat/cleanup
+          artifacts:
+            - behat-screenshots/**
+
 image: quay.io/pantheon-public/build-tools-ci:5.x
```
aaronbauman: docker image 6.x has a compatibility bug https://github.com/pantheon-systems/docker-build-tools-ci/issues/36

aaronbauman:
All these steps are passing up until Behat.
Not sure why behat is failing, and pipelines seems to have a bug that it won't allow download of artifacts, making it very difficult to troubleshoot.

Here's the test project I'm working on: https://bitbucket.org/messageagency/pipe-test/addon/pipelines/home#!/results/%7Bf87b1397-a8b5-42bb-a0f1-7e89d54fd91a%7D/steps/%7B2cea18cf-2b6e-43ab-8627-bbfb0a4e5fd7%7D

bitbucket-pipelines.yml
ataylorme: Yes, Behat changes could make visual tests fail. For eaxample, a new post would change latest posts and archive templates. 

The current Behat tests in this example change site name and description, which are visible on all pages. If Backstop runs after Behat makes those updates the comparison of the multidev to dev will fail.

I would expect those who don't need/want tests, such as Behat, to remove them.

bitbucket-pipelines.yml
ataylorme: > using local set-environment script here, so that we can run build tools dev-master on top of arbitrary docker image

Is that what you are doing or are you suggesting a change here?

bitbucket-pipelines.yml
ataylorme: That bug was resolved. Are you still experiencing it?

ataylorme:
> All these steps are passing up until Behat.
Not sure why behat is failing

@aaronbauman looking at your log it seems you are running this on D8, not WordPress. Try persisting the `drush` directory, along with `web` and `vendor`, from the PHP build job.

Without it, the Drush Behat extension won't be deployed to the Pantheon site and functionality (creating `nodes` and `terms`) will fail.

bitbucket-pipelines.yml
aaronbauman: Yes, still getting this error, even on the bleeding-build-tools tag.
I just commented on the other thread.

bitbucket-pipelines.yml
aaronbauman: I would rather not maintain this set-environment file separately, but currently it's the only way to install build tools dev-master and work around bugs in docker image.

One of these needs to be resolved before we can get rid of set-environment from this repo:

  -  https://github.com/pantheon-systems/docker-build-tools-ci/pull/41  
  - https://github.com/pantheon-systems/docker-build-tools-ci/issues/36

aaronbauman:
Thanks for all that feedback.
Latest commit builds successfully!

ataylorme:
@aaronbauman just an FYI I opened #278 and am testing it now with Build Tools beta 16 by:

- creating a new project with `build:project:create`
- creating a new branch/PR
- merging the branch/PR

The only change I made from what I have in the WordPress version is persisting the `drush` directory on the PHP build step. I'll let you know how it goes.

bitbucket-pipelines.yml
```diff
@@ -1,27 +1,122 @@
...
-          - composer
-        script:
-          - apt-get update && apt-get install -y unzip
-          - curl -sS https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
-          - composer install --no-ansi --no-interaction --optimize-autoloader --no-progress
-          - CI_BRANCH=$BITBUCKET_BRANCH CI_BUILD_NUMBER=$BITBUCKET_BUILD_NUMBER PR_NUMBER=$BITBUCKET_PR_ID /build-tools-ci/scripts/set-environment
-          - source ~/.bashrc && composer -n install
+definitions:
+  caches:
+    terminusplugins: ${TERMINUS_PLUGINS_DIR:-~/.terminus/plugins}
```
ataylorme: @aaronbauman what's the reason for caching Terminus plugins? They are pre-installed in the Docker image

aaronbauman: debugging cruft

aaronbauman:
Closing in favor of #278

PR #283, opened 2019-09-20
Proof of concept scheduled composer lock update job using Terminus CLU plugin

aaronbauman:
I put together this [Terminus CLU plugin](https://github.com/aaronbauman/terminus-clu-plugin), which relies on Build Tools APIs, and generally seems a bit more elegant than the existing clu command.

I've got this working on some test projects right now, but wanted to get it on your radar to get some feedback.

Still todo: create the scheduled job during project creation.

### DIFF
--- a/bitbucket-pipelines.yml
+++ b/bitbucket-pipelines.yml
@@ -111,3 +111,19 @@ pipelines:
   branches:
     # Using the default_steps anchor means no other items can be added
     master: *default_steps
+  custom:
+    clu:
+    - step:
+        name: Composer Lock Update Test
+        services:
+          - docker
+        caches:
+          - docker
+          - composer
+        script:
+          - *bash_env_export
+          - *bash_env_source
+          - /build-tools-ci/scripts/set-environment
+          - git clone https://github.com/aaronbauman/terminus-clu-plugin.git ${TERMINUS_PLUGINS_DIR:-~/.terminus/plugins}/terminus-clu-plugin
+          - terminus -n auth:login --machine-token="$TERMINUS_TOKEN"
+          - terminus clu $BITBUCKET_REPO_SLUG


aaronbauman:
The CLU features could be simplified quite a bit if they were merged directly into Build Tools plugin. Let me know if that's something maintainers would consider, and I'll create a PR against Build Tools.

ataylorme:
> Still todo: create the scheduled job during project creation.

That can be done by updating the CI config files in `example-drops-8-composer` and `example-wordpress-composer`.

> The CLU features could be simplified quite a bit if they were merged directly into Build Tools plugin. Let me know if that's something maintainers would consider, and I'll create a PR against Build Tools.

I am in favor of this but will discuss with @greg-1-anderson as well. @danielbachhuber what are your thoughts? I know you've expressed interest in letting Pantheon maintain CLU, and this seems like a good approach.

@aaronbauman it would also be interesting to make CLU in build tools more agnostic, starting with support for `composer.lock` as well as `yarn.lock`/`package-lock.json`.

danielbachhuber:
I _think_ my preference would be to continue to keep CLU separate. It has value as a standalone utility, in the same way it doesn't makes sense to bundle into WP-CLI.

But, not a strong opinion. Ideally, any sort of moving would require adding tests, because we've been talking about adding tests forever (https://github.com/danielbachhuber/composer-lock-updater/issues/3)

aaronbauman:
My biggest motivation for creating a terminus plugin was to re-use the existing API clients and authentication.

Obviously `clu` has value on its own, outside a terminus or pantheon project... Which got me thinking, what if those clients were abstracted into their own standalone PHP libraries? Then clu and terminus and various plugins could leverage those, instead of all the overlapping concerns... But I don't want to go off on too much of a tangent here...

ataylorme:
> Which got me thinking, what if those clients were abstracted into their own standalone PHP libraries? Then clu and terminus and various plugins could leverage those, instead of all the overlapping concerns... But I don't want to go off on too much of a tangent here

That is an interesting thought. I'm not sure if such a change could be done in a non-breaking way. WP-CLI moved to a package approach and would be good inspiration.

Lots of folks are currently relying on the (unstable) `2.x` version of Build Tools and such a big change feels like a good initiative to drive a version `3` of Build Tools but I'm very much open to discussing the idea further.

greg-1-anderson:
Yeah we should make a stable release of 2.

I am a little reluctant to roll CLU into this project. Going the other way as suggested seems better. I have broken out some things, like hubph, but more refactoring would be useful (hubph does not to gitlab / bitbucket, for example)

ataylorme:
@aaronbauman thank you for your work here! Would you object to https://github.com/aaronbauman/terminus-clu-plugin being moved to `pantheon-systems` and becoming a Pantheon maintained plugin?

That would shift the commitment of maintaining the plugin from you to Pantheon. It would also mean we can add it to https://github.com/pantheon-systems/docker-build-tools-ci/ and simplify this script a bit

aaronbauman:
Moving to pantheon-systems would be fine by me, if you're interested in maintaining it. The commands would also be simpler if they were rolled into Build Tools Plugin, which would allow for getting rid of the wrapper layer.

As for this specific PR: adding CLU bitbucket-pipelines.yml only exposes the job to BB's scheduled task API. As yet, the scheduled job must be created manually.

ataylorme:
> As yet, the scheduled job must be created manually.

Seems like we need a corresponding issue/PR to set the scheduled job in Pipelines during `build:project:create`. We are already doing something similar for GitLab.

However, I think we should land this PR first so that we have a finalized job name to use in that work.

ataylorme:
The `deploy_to_pantheon` job here is failing because `TERMINUS_TOKEN` is not set. I believe that is because the PR came from a fork. I re-created the changes in #289 to see if tests will pass there.

I'll also test the CLU job manually in a BitBucket project

ataylorme:
> Seems like we need a corresponding issue/PR to set the scheduled job in Pipelines during `build:project:create`. We are already doing something similar for GitLab.

Here is the API endpoint for that: https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Bworkspace%7D/%7Brepo_slug%7D/pipelines_config/schedules/

ataylorme:
@aaronbauman this is working well in my tests (after BitBucket recovered from its outage). https://bitbucket.org/ataylorme/example-d8-composer-pipelines-test-2019-10-01/pull-requests/1/update-composer-dependencies-2019-10-02-17/diff was opened

aaronbauman:
If someone can help me make sense of the bitbucket API, I can write a PR for adding the scheduled task. I've tried POSTing [several different data structs](https://community.atlassian.com/t5/Bitbucket-questions/Need-help-posting-to-bitbucket-api-endpoint/qaq-p/1193244), but haven't found any that work.

ataylorme:
> If someone can help me make sense of the bitbucket API

Yeah, their docs aren't super clear.

`cron_patternstring: The cron expression that the schedule applies.` is pretty vague.

ataylorme:
@aaronbauman FYI I started https://github.com/pantheon-systems/terminus-clu-plugin

ataylorme:
Closing in favor of #289

PR #291, opened 2019-10-02
Use "-r" instead of "-a" in rsync

### LINKED ISSUE #184: Remove Quicksilver deploy_product hook
I would like to remove the `deploy_product` key from `pantheon.yml`. This section was added to accommodate the use case of a custom upstream not committing a vendor directory and relying on the site spin-up process to do the composer install.

https://github.com/pantheon-systems/example-drops-8-composer/blob/master/pantheon.yml#L5

I don't know if any upstream users ever did this and I would not recommend anyone do it going forward. @greg-1-anderson any objection to removing?

nelsonamaya82:
This fixes #184 by replacing -a flag for -r flag in the rsync commands (special thanks to @ataylorme).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,7 +138,7 @@ jobs:
       # Rsync the build artifact pieces web directory
       - run:
           name: sync build artifact
-          command: rsync -az /tmp/web /tmp/vendor /tmp/drush .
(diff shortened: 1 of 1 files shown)


ataylorme:
Tests running in #292

PR #298, opened 2019-10-17
drupal-scaffold mention

alisonjo315:
see also: https://github.com/pantheon-systems/drops-8-scaffolding/issues/2


greg-1-anderson:
This advice was useful; I am sorry I never merged it. It is obsolete now; with the move to drupal/core-composer-scaffold, scaffold assets are now handled differently than described here.

alisonjo315:
Oh thanks for the feedback! No worries :)

PR #311, opened 2019-11-22
Diffy visual regression testing integration

ygerasimov:
Adds an integration with https://diffy.website. It was built with an example of BackstopJS integration (same changes/commit command triggers the job).

Also terminus command https://github.com/DiffyWebsite/diffy-build-tools-plugin is available to make it easier configure CircleCI.

### DIFF
--- a/.ci/test/diffy-visual-regression/README.md
+++ b/.ci/test/diffy-visual-regression/README.md
@@ -1,0 +1,7 @@
+# Diffy Integration
+
+You need to set up two environmental variables `DIFFY_API_KEY` and `DIFFY_PROJECT_ID` in CircleCI for integration to work.
+
+There is a terminus command [https://github.com/DiffyWebsite/diffy-build-tools-plugin](https://github.com/DiffyWebsite/diffy-build-tools-plugin) available to help you out.
+
+Read more about this set up at [https://diffy.website/documentation/pantheon-build-tools](https://diffy.website/documentation/pantheon-build-tools).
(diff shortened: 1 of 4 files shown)


stevector:
Hi @ygerasimov, thanks for showing that this is possible! But I'm reluctant to add this kind of feature to this example. Just yesterday I was chatting with co-workers in our internal Slack about my concern that the size and complexity of this repo makes it difficult for people to adopt if they're first just looking to adopt Composer and a PR workflow. Adding a second visual regression testing service when Backstop is already present is more complexity than benefit IMO.

@greg-1-anderson, I'm curious what you think here too. I'd prefer that Diffy was document as a potential integration, maybe through a CircleCI Orb, rather than something baked into this example repo.

ygerasimov:
@stevector I completely understand your concern. For sure having two visual regression services doesn't make sense.

I was thinking about a set up like you have in quicksilver examples https://github.com/pantheon-systems/quicksilver-examples, where you could have multiple solutions available for you. So it you simply pick an choose the ones you like.

As for Diffy, maybe I could comment out a diffy_visual_regression_testing workflow job in config.yml and provide instructions how to set it up. Then by default it won't be applicable, but still people could discover the possibility and use it easily if needed? What do you think?

stevector:
@ygerasimov, yeah, I'd be fine with with adding something like a heading of "Additional Testing Examples" to the readme of this project that could list integrations like Diffy.  I think CircleCI Orbs are the best available mechanism for distributing examples like that.

I do not want to go the route of a commented-out example directly in `config.yml` because that file is currently 269 lines long which is already in the range to overwhelm a CI newcomer.

PR #319, opened 2019-12-09
Update composer.json

ChristopherMatthews:
Git sub-modules can easily confused new users to Composer. Let's make it not quite a painful at least on the Pantheon platform. I'm not 100% sure this is the correct approach, but if the removal of git sub-modules can be automated I think it would be a win.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -69,14 +69,20 @@
         "prepare-for-pantheon": "DrupalProject\\composer\\ScriptHandler::prepareForPantheon",
         "post-install-cmd": [
             "@drupal-scaffold",
-            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles",
+            "find docroot vendor -name '.git' | xargs rm -rf",
+            "find docroot vendor -name '.gitignore' | xargs rm -rf"
         ],
         "post-update-cmd": [
-            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles"
+            "DrupalProject\\composer\\ScriptHandler::createRequiredFiles",
+            "find docroot vendor -name '.git' | xargs rm -rf",
+            "find docroot vendor -name '.gitignore' | xargs rm -rf"
         ],
         "post-create-project-cmd": [
             "@drupal-scaffold",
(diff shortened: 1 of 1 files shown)


stevector:
Hi @ChristopherMatthews, thanks for the PR! We also have this need addressed in the Terminus Build Tools Plugin, which is meant to be used with this example repo: https://github.com/pantheon-systems/terminus-build-tools-plugin/blob/master/src/Commands/BuildToolsBase.php#L787

I'd like @greg-1-anderson and @sugaroverflow's input on whether we should duplicate the functionality. I would prefer not to duplicate it, but if we did, I would rather see this in a Composer plugin that this example project, or others, could `composer require`. I'm reluctant to increase the amount of custom scripts in this repo that a developer might need to copy into a new or pre-existing projects that they are migrating to this structure.

sugaroverflow:
👋 hello @ChristopherMatthews, thank you for working on a PR for this!

I agree with @stevector and would prefer not to duplicate this functionality here. I've seen two ways of solving this issue in the past and believe that it should be something optional that a developer can require as needed because you can run into issues when dealing with development modules or updating dependencies (that previously had git submodules removed).

In the past, I've used [a composer script solution similar to yours](https://github.com/drupal-composer/drupal-project/issues/223#issuecomment-266417254) - the structure is slightly different as it uses a script file + functions to centralize the removal code.

And also [composer-cleanup-vcs-dirs](https://github.com/VolantisDev/composer-cleanup-vcs-dirs) performs a similar function when included in a `composer.json`. I'd have to dig a bit into it to see how this differs from the script. 

Doing some research, it appears a lot of drupal sites are using a similar script and/or patch their site with similar code - so perhaps it would be benefit from being a plugin of its own that can be included.

greg-1-anderson:
We used to have a script called "Prepare for Pantheon" that did this operation. Now it is done as part of `build:env:create` and `build:env:push` in the Terminus build tools.

It might make sense to update the [No CI Workflow](https://pantheon.io/docs/guides/drupal-8-composer-no-ci) document to use `build:env:push` instead of `git push --force`. That would address this need without duplication.

greg-1-anderson:
Let's track this in the above-referenced documentation update.

PR #322, opened 2019-12-11
We do not need both drupal/core and drupal/core-recommended

greg-1-anderson:


greg-1-anderson:
The problems here were cleared up when we switched to drupal/core-composer-scaffold

PR #326, opened 2019-12-20
Update settings.php to latest Drupal 8.8.0 recommendations.

alisonjo315:
Use $settings['config_sync_directory'] instead of $config_directories[CONFIG_SYNC_DIRECTORY].

Reference: https://git.drupalcode.org/project/drupal/commit/1a2babd20df4af74961fbb043265a1122aa822f9#b2f0ac5b08b48f86f4dc3443ff06401953a4c17a

PR #327, opened 2019-12-20
Update settings.php to latest Drupal 8.8.0 recommendations.

greg-1-anderson:
Use $settings['config_sync_directory'] instead of $config_directories[CONFIG_SYNC_DIRECTORY].

From #326


greg-1-anderson:
Hm I thought this would apply the credentials so that the tests could run, since I pushed to origin.

greg-1-anderson:
Ah, good, that did work after all.

PR #355, opened 2020-03-07
Update composer.json

ChristopherMatthews:
Update drush/drush to ~10.2.2
https://packagist.org/packages/drush/drush#10.2.2
https://github.com/drush-ops/drush/releases/tag/10.2.2


greg-1-anderson:
Sure, we could do that. However, we should use `~10` or `^10.2.2`; there is no reason to exclude 10.3 and later here.

PR #404, opened 2020-07-16
Add HTTP Basic Auth for locked environments

jameswilson:
Documents the fix in PR pantheon-systems/docker-build-tools-ci#60 for issue pantheon-systems/docker-build-tools-ci#59


greg-1-anderson:
Looks good to me.

If folks want to automate the password lookup via Terminus, I think it would be best to use the CircleCI API to set the environment variable after looking it up (e.g. in an earlier CircleCI job). See the Terminus Build Tools plugin for code to set CircleCI environment variables.

n.b. The username and password are hidden in the output because they are set from CircleCI environment variables. My earlier suggestion to have a hook to call Terminus from the setup scripts would lose this protection.

PR #468, opened 2021-05-18
Add GitHub Actions workflows files.

kporras07:
Github Actions initial support added.

Several secrets should be added to Github Actions configuration to make it work:
- ADMIN_EMAIL
- ADMIN_PASSWORD
- ADMIN_USERNAME
- TERMINUS_TOKEN
- TERMINUS_SITE
- SSH_PRIVATE_KEY
- GH_TOKEN

I'm waiting for [this PR](https://github.com/garris/BackstopJS/pull/1317) to avoid compiling and updating git in runtime.


greg-1-anderson:
So, something I just thought of: if we merge this PR, then GitHub is going to try to run these tests on every PR made in this repository, right?  We don't want that. Maybe we should store the GitHub actions templates at a different path, and move it into position in the Terminus Build tools. There should also be a README with instructions for anyone working on it manually.

kporras07:
@greg-1-anderson you're right. I just updated the files location. I'll do the same in the other PRs. Thanks

kporras07:
I changed the way of updating git so that instead of taking ~2.5 min, it now takes ~10s.

README.md
```diff
@@ -1,116 +1,5 @@
-# Example Drops 8 Composer
+# drop8kporras07
```
greg-1-anderson, line 1: I think the changes to this file were unintentional

greg-1-anderson (changes requested):
Just needs README fixed up, should be similar to https://github.com/pantheon-systems/example-drops-7-composer/pull/27/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

kporras07:
@greg-1-anderson fixed by rebasing. Thanks.

I had messed up with the repo when testing build-tools integration.

